PR #135, opened 2026-06-23
Recover from transient stream failures: provider auto-retry + error/retry UI

oratis:
## What & why

Diagnosed a local Lisa failure: a chat turn died with `[error] request ended without sending any chunks` (twice), and the same error hit `[idle]` reflection in `backend.log`. The Anthropic SDK throws this when a streaming response opens (HTTP 200) but the SSE body closes before any event arrives — here, the Clash/mihomo proxy (`127.0.0.1:7897`) tearing down an idle CONNECT tunnel during a long time-to-first-byte. The SDK's own retries don't cover it (thrown while iterating a stream that already "succeeded"), so a momentary blip became a hard, user-visible error with no recovery.

Two complementary layers of fix:

### 1. Provider auto-retry (silent, proxy-agnostic) — `fix(providers)`
- New shared `withStreamRetry()` ([stream-retry.ts](src/providers/stream-retry.ts)) wired into **all three** providers (anthropic / openai / gemini).
- Retries the empty-stream error + connection resets / premature closes, **only while no delta has been forwarded yet** — so already-streamed output is never duplicated, and a user abort is never retried.

### 2. Error detail + retry button (manual fallback) — `feat(web)`
- Replaces the bare `[error] …` line with an `.err-block`: error detail + **↻ 重试** button that re-runs the same message + files. Covers what auto-retry can't (non-retryable errors, or retries exhausted).
- **Fixes double-rendered error**: `runAgent` emits an error event *and* rethrows, and the server catch sent a second identical event (this is why the screenshot showed it twice). Deduped server-side (`errorSent`) and client-side.
- **Fixes retry duplicating the user message**: the user message was persisted *before* the provider call, so an immediate stream failure orphaned it on disk and a retry wrote it again. Now persisted lazily, only after the first provider response commits the turn.

## Testing
- New: `withStreamRetry` / `isRetryableStreamError` unit tests, Anthropic provider retry integration tests, and a `runAgent` test asserting zero persistence when the first provider call throws.
- Existing provider/agent persistence-order tests unchanged and green.
- `html-syntax.test.ts` confirms the inline `<script>` still parses; snapshot recomputed.
- Full suite: **829 pass / 1 skip / 1 fail**. The lone failure is `src/cli/pair.test.ts`, a pre-existing `qrcode-terminal` module-not-found in this checkout — zero overlap with these changes.

> Note: environment-side, pinning `api.anthropic.com` to a stable Clash node (vs url-test/fallback flapping) reduces the transient drops at the source; this PR makes Lisa resilient to them regardless.

### DIFF
--- a/docs/PLAN_MAIL_v1.0.md
+++ b/docs/PLAN_MAIL_v1.0.md
@@ -1,0 +1,203 @@
+# LISA Mail Module — Product Design (v1.0, draft)
+
+> Goal: connect the user's real mailboxes; once a day LISA reads everything,
+> **classifies + grades** it, and pushes a **daily digest**; anything truly
+> important triggers a **proactive chat message + push**. Read-first, privacy-first.
+
+## 0 · Reference: agent.qq.com (Tencent "Agently Mail")
+
+What it actually is (researched 2026-06): a **dedicated, isolated mailbox _for_ an
+AI agent** — a separate address the agent owns, walled off from your personal QQ/
+WeChat mail, so the AI only ever touches *its* inbox. Highlights: **two-stage write
+confirmation** (send/reply/forward/delete → generate a summary → user confirms →
+execute), **prompt-injection defense** on read, **A2A** (agent-to-agent: inter-company
+quote/order automation), and packaged scenarios (invoice→reimbursement, subscription
+digest, AI resume submission, order reconciliation).
+
+**Our model is different** — LISA reads *your existing* inboxes (Gmail/QQ/IMAP) and
+acts as an inbox chief-of-staff. But three ideas transfer directly:
+1. **Two-stage confirmation for any write** → reuse LISA's existing per-action
+   approval gate (the one managed agents already use). Read-only in v1 regardless.
+2. **Treat email bodies as untrusted input** → prompt-injection hardening when the
+   classifier reads them.
+3. **Scenario taxonomy** (invoice/receipt, newsletter digest, order, recruiting) →
+   seeds our classification categories. A2A (LISA ↔ other agents' mailboxes) is a
+   natural later tie-in to the existing `takoapi` gateway.
+
+## 1 · One-liner
+
+> Connect your mailboxes once. Every morning LISA hands you a single ranked digest
+> instead of an inbox — and taps you the moment something actually needs you.
+
+## 2 · The three capabilities (the spec)
+
+| # | Capability | Outcome |
+|---|---|---|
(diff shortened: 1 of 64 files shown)


oratis:
## Review — provider auto-retry + error/retry UI

Solid, well-tested change. The core invariant — **only retry while no delta has been emitted** — is the right safety model, and `markEmitted()` is threaded correctly through all three providers. Abort handling is careful (name-based `APIUserAbortError` / `AbortError` exclusion + `signal.aborted` guard + abortable backoff delay), and the test matrix is genuinely thorough: retryable/non-retryable classification, no-retry-after-emit, max-retries exhaustion, pre-aborted signal, provider integration, and the `runAgent` zero-persistence test. The two incidental bug fixes (server-side `errorSent` dedup; deferred user-message persistence to avoid the orphan/duplicate on retry) are correct, and error detail is rendered with `textContent` (XSS-safe). Dropping the dead `finish` var in Gemini is a nice cleanup.

### Blocking before merge
- ⚠️ **Stale / conflicting.** Branch is **27 commits behind `main`**; GitHub reports `CONFLICTING`. All four web files it touches (`lisa-client.ts`, `lisa-css.ts`, `server.ts`, `lisa-html-snapshot.test.ts`) changed on `main` since the branch point (console redesign + sidebar rework). Concretely the snapshot here is `EXPECTED_LENGTH=95307` but `main` is now `143704`, so the byte-exact snapshot test will fail post-merge and must be recomputed. The provider layer (`stream-retry.ts`, `anthropic/openai/gemini.ts`, `agent.ts`) is outside the conflict surface and reconciles cleanly; the conflict is entirely in the web-UI layer — and note `main`'s `send()` composer path was refactored, so `.err-block` must be re-applied onto the new flow, not the old diff.

### Non-blocking suggestions
- `isRetryableStreamError`: `msg.includes("terminated")` is broad — it correctly catches undici's `TypeError: terminated`, but also matches any API error message containing "terminated". A false positive triggers a silent re-send (extra latency + re-billed input tokens), so prefer `err.name` / `err.cause?.code` where available.
- Each retry reopens a fresh request, so input tokens are re-billed per attempt (bounded by `maxRetries=2`). Worth a one-line comment to make it a conscious tradeoff.
- `throw lastErr` at the end of `withStreamRetry` is unreachable (the `i === maxRetries` branch always throws first) — harmless; it satisfies the type checker.

Addressing all of the above now (reconcile onto current `main`, re-apply the err-block to the refactored composer, tighten the retryable matcher, recompute the snapshot, re-run the suite) before merging.

PR #142, opened 2026-06-25
feat(ui): agent operations console — clients + web redesign

oratis:
## What

Redesigns the LISA **client and web interface** into an "agent operations console" (per a reference layout), mapped onto LISA's **real data** — no fictional OKR backend. **Chat stays the default home**; Dashboard / Control / Rêve / Sense / Memory are new nav destinations.

## Surfaces

- **Backend** — the one missing primitive: a real proactive/autonomy master switch. New `~/.lisa/autonomy/state.json` + `GET/POST /api/autonomy/state` (token-gated, so the phone can flip it); the `idle` and `heartbeat` runners no-op when it's off.
- **Web / Mac** (the Mac app is a WKWebView wrapper, so it inherits this) — left rail (workspace pill · identity · nav list · live agents card · PROACTIVE toggle) + a main view-stack. The chat/SSE/birth/cfg ids are byte-intact; everything is **additive**. Five new views pull live data; the byte-exact `MAIN_HTML` snapshot is regenerated.
- **iOS** — shared `Theme.swift`, global dark + cyan tint, restyled 5 tabs, a green Proactive banner + stat strip on Dispatch, and a mirrored toggle in Settings.
- **Island** — green accent token only (backtick-safe).

## Mapping (reference → LISA)

| Reference | LISA |
|---|---|
| Departments | live agent sessions |
| OKR stat bar | Agents · Tasks · Signals |
| Objectives card | Lisa's current desire / focus |
| Key Results cards | agent & dispatch-task cards |
| PROACTIVE toggle | autonomy (idle + heartbeat) |

## Verification

- `npm run typecheck` ✅ · **828 tests, 0 failures** (snapshot regenerated, inline-JS syntax guards green).
- Web run + screenshots: chat is default & unchanged, all views render real data, and the proactive toggle round-trips to the server (persists `enabled` to `~/.lisa/autonomy/state.json`).
- iOS: `swiftc -parse` clean on all files; full compile is CI/Xcode (no Xcode in this environment).

## Notes

- Nav is a **vertical list inside the sidebar** (matches the approved mockup / reference) rather than a separate icon rail, so the `.frame` grid is unchanged — lower risk.
- The Control view ships approve/deny/cancel + PTY output; per-row "send" was left out to avoid guessing its request shape.
- A self-contained click-through mockup exists locally at `mockups/lisa-console.html` (not committed — fake data + a throwaway preview server).

### DIFF
--- a/packaging/ios-companion/Sources/App.swift
+++ b/packaging/ios-companion/Sources/App.swift
@@ -28,6 +28,10 @@ struct RootView: View {
             SettingsView()
                 .tabItem { Label("Settings", systemImage: "gearshape") }.tag(4)
         }
+        .tint(Theme.accent)                                  // cyan active tab + links + controls
+        .preferredColorScheme(.dark)                         // force the console dark look
+        .toolbarBackground(Theme.panel, for: .tabBar)
+        .toolbarBackground(.visible, for: .tabBar)
         .onOpenURL { app.handleDeepLink($0) }
         .overlay { if app.locked { LockView() } }
         .onChange(of: scenePhase) { _, phase in
--- a/packaging/ios-companion/Sources/AppState.swift
+++ b/packaging/ios-companion/Sources/AppState.swift
@@ -149,4 +149,37 @@ final class AppState: ObservableObject {
             locked = false
         }
     }
+
+    // ── proactive mode (autonomy on/off), mirrored from /api/autonomy/state ──
+    /// Whether Lisa self-drives (idle + heartbeat) when you're away.
+    @Published var proactiveEnabled = true
+    /// Disables the toggle mid-flight so a double-tap can't race.
+    @Published var proactiveBusy = false
+    /// False when the Mac is too old to expose /api/autonomy/state (404) — the
+    /// toggle then renders disabled rather than lying about a state it can't set.
+    @Published var proactiveAvailable = true
+
+    func loadProactive() async {
+        do {
+            proactiveEnabled = try await client.autonomyState()
+            proactiveAvailable = true
+        } catch LisaError.http(404) {
+            proactiveAvailable = false
+        } catch {
+            // transient (offline / timeout) — keep the last-known state, don't flip the UI
+        }
+    }
+
+    /// Optimistic flip with rollback on failure (mirrors how fire() actions behave).
+    func setProactive(_ on: Bool) {
+        guard !proactiveBusy else { return }
+        let previous = proactiveEnabled
+        proactiveEnabled = on
+        proactiveBusy = true
+        Task { @MainActor in
+            defer { proactiveBusy = false }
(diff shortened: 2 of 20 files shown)


oratis:
## Review — approve ✅ (cleanups applied in 10aca25)

Reviewed the full diff and verified the risky bits against the actual code rather than the description.

**Independently verified on the branch**

| Check | Result |
|---|---|
| `tsc --noEmit` | ✅ clean |
| full suite | ✅ **828 tests, 0 failures** (1 expected PTY/Node-26 skip) |
| `lisa-html-snapshot` (length + sha256) | ✅ byte-exact, regenerated |
| `/api/autonomy/state` behind the 401 auth gate | ✅ gate precedes handlers |
| idle no-op return matches `IdleRunResult` (5 fields) | ✅ exact |
| `escapeHtml` in scope for console JS | ✅ (+ guarded) |
| `AppState` is `@MainActor` (post-`await` `@Published` writes safe) | ✅ |
| all 9 endpoints the new views fetch exist server-side | ✅ |
| `swiftc -parse` on touched iOS sources | ✅ |

**Strengths:** `autonomy/state.ts` mirrors `control/policy.ts` (pure normalize, corrupt-file tolerance, lazy `statePath()`, default-on preserves historical behavior); both runners gate *before* taking their locks and the watcher keeps ticking so re-enabling resumes immediately; the web change is genuinely additive (chat IDs/grid preserved, views lazy-built) with the byte-exact snapshot regenerated rather than dropped; toggle has proper `role="switch"`/keyboard a11y and optimistic-with-rollback on both clients.

**Cleanups applied (10aca25):**
1. The 60s `refreshClaudeSessions` tick captured the **original** fn by reference before `setupConsole` wrapped it, so console views/nav-count only refreshed on SSE events. Now resolves `window.refreshClaudeSessions` at call time → the wrapper wins on the periodic tick too.
2. Removed the now-orphaned iOS `struct Pill` (all call sites migrated to `ThemePill`).
3. Removed the unused `.pp-tag.you` CSS rule.
4. Regenerated the `MAIN_HTML` snapshot (len 130688) for the two web changes.

**Considered but intentionally kept:** `POST /api/autonomy/state` uses `saveAutonomyState({ ...loadAutonomyState(), ...payload })`. On a second look this isn't dead — it's preserve-on-partial-POST (and forward-compat once `AutonomyState` grows a second field); simplifying to `payload` would make `POST {}` reset to the default `enabled:true`. Left as-is.

**Note (pre-existing, not this PR):** the new POST inherits the whole server's "loopback ⇒ trusted, no token" posture, same as `/chat`. A remote *authed* device can also toggle autonomy (deliberate, per the inline comment) — strictly less powerful than the `/chat` it can already drive. No new exposure.

Merging.

PR #167, opened 2026-06-26
feat(ios): first-run onboarding — install LISA on your Mac & pair (M1)

oratis:
**Note:** the M1 onboarding feature itself already reached `main` via #168 (the bundle-id PR was stacked on this branch and carried these files in). To avoid re-merging identical code as an add/add conflict, this branch was rebuilt on top of `main` and now contains **only the follow-up polish** that wasn't in #168:

1. **Plan ↔ impl reconcile** — `PLAN_IOS_ONBOARDING_v1.0.md` decision ① said the LISA Cloud card was a disabled "Coming soon" state; the shipped `OnboardingFlow` actually makes it *active* (manual `https://…?token=` paste via `parsePairing`), with only Sign in with Apple deferred to M4. Doc now matches code. Also records *why* decision ② binds `0.0.0.0` always rather than gating on first-pair (pairing is loopback-only, so gate-on-pair is feasible + more least-privilege, but costs a backend restart mid-pair).
2. **UX** — the "Can't reach your Mac" help now mentions the macOS Application Firewall "allow incoming connections" prompt, the silent-failure mode of the default `0.0.0.0` bind (companion to #169).

## Verification
- `packaging/ios-companion/build.sh` → **BUILD SUCCEEDED** (iPhone 17 Pro sim).

### DIFF
--- a/docs/PLAN_IOS_ONBOARDING_v1.0.md
+++ b/docs/PLAN_IOS_ONBOARDING_v1.0.md
@@ -33,14 +33,27 @@ default-bind tweak (decision ②).
 ## Decisions (locked)
 
 1. **Ship the My Mac / LISA Cloud fork now** (flow step 1). The Mac path is fully
-   functional; the Cloud card routes to the `edition.ts` sign-in if ready, else a
-   "Coming soon" state. A single `ConnectionMode` threads through later steps and
-   drives `parsePairing`'s `http` (Mac) vs `https` (cloud) scheme.
+   functional. **The Cloud card is active in M1** — it routes to manual paste of an
+   `https://…?token=` cloud URL (the same `parsePairing` path), so a user with a
+   LISA Cloud endpoint can connect today; only *Sign in with Apple* is deferred
+   (shown disabled, "coming soon", wired in M4). A single `ConnectionMode` threads
+   through later steps and drives `parsePairing`'s `http` (Mac) vs `https` (cloud)
+   scheme.
 2. **The menu-bar Mac app binds LAN-reachable by default** (`--host 0.0.0.0`).
    This removes the #1 failure (forgetting the flag) for app users and lets the
    flow's "Start LISA" step branch by install method. LAN-reachable is still
    **token-gated** — someone on your Wi-Fi can attempt to connect but can't
    without the paired device token; the flow says so on-screen.
+   - *Why always-on, not "bind 0.0.0.0 only once a device is paired":* pairing is
+     loopback-only (the QR is minted on the Mac), so gate-on-first-pair is feasible
+     and is strictly more least-privilege — but it costs a backend restart at the
+     exact moment the user is mid-pair, the most failure-sensitive step. Always-on
+     keeps the port open behind the token gate; the residual exposure (a future
+     auth regression / DoS / "this Mac runs LISA" fingerprinting) is the accepted
+     trade for a restart-free pair. Revisit if we ever bind without a token.
+   - *Firewall:* the first `0.0.0.0` bind can trip the macOS Application Firewall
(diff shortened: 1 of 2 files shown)


oratis:
## Review — solid M1, ready to merge

Pure iOS UX over the existing pairing primitives with no new backend protocol, exactly as the plan promised. I verified the parts most likely to be subtly wrong:

- **"Lands on Chat" is correct.** `finishOnboarding(paired: true)` sets `selectedTab = 1`, and tab 1 is `ChatView` (`App.swift:23`); tab 0 is Dispatch. No off-by-one. ✅
- **Install commands match the README exactly** — `npm install -g @oratis/lisa` (README:139), `brew install oratis/tap/lisa` (README:146), and the `Lisa-Suite.dmg` → `releases/latest` link (README:132). The tests pin these, so drift will be caught. ✅
- **`verifyConnection`'s `404 → .ok` mapping is sound.** The auth gate (`server.ts:568–583`) runs *before* route dispatch, so an unauthorized request is always `401`, never `404`. A `404` therefore genuinely means "reachable + token accepted, just an older Mac without the island-ping endpoint." Good reasoning. ✅

### Substantive
1. **The Cloud card diverges from the locked plan.** Decision #1 and the flow table in `PLAN_IOS_ONBOARDING_v1.0.md` say the LISA Cloud card shows a disabled **"Coming soon"** state until the cloud edition lands. The implementation makes it *active* and routes to manual paste of an `https://…?token=` URL (only Sign in with Apple is deferred). That's arguably a better M1 (functional now), but doc and code now disagree — please update the plan, otherwise someone "fixes" the card back to disabled later.
2. **"I already have LISA running" skips the mode fork.** `welcome → go(.pair)` leaves `connectionMode` at its prior value, and `pair`'s manual entry defaults to `.mac`. A *returning cloud* user — arguably the most likely presser of "I already have it" — lands in the Mac pairing screen. Fine as an M1 escape hatch, but consider routing it through `.mode` first, or having the pair-step manual entry honor the existing `connectionMode`.

### Minor
3. The "Download Lisa-Suite.dmg" button links to the releases *listing* page, not the asset directly (same as the README, so consistent) — the label promises a specific file. Either acceptable.
4. `showOnboarding` is computed once in `init()` from `!isConfigured && !onboarded`. The `needsSetup` banner covers the re-entry paths, so this is fine — just noting it's not reactive to later config changes within a launch.
5. Tests are pure-model only (the noted headless-sim limitation) — the step machine and the `verifyConnection` outcome mapping aren't exercised. Acceptable for M1; worth revisiting when the sim runner is available.

oratis:
### Review — first-run onboarding (M1)

⚠️ **Process finding worth flagging:** the entire M1 implementation **already landed on `main`** via commit `5a97597` (#168, titled *"fix(ios): bundle id → ai.meetlisa.main"*). That commit actually carried 895 insertions including every `Sources/Onboarding/*` file, the `AppState`/`App`/`SettingsView` hooks, and the tests. That's why this PR shows `CONFLICTING` — git sees add/add on the files #168 already added.

So the **real delta of this PR vs `main`** is just:
- the extra firewall hint line in `failureView` (*"If macOS asked to allow incoming connections, choose Allow."*) — a genuine improvement, and
- the expanded rationale in `docs/PLAN_IOS_ONBOARDING_v1.0.md` (the always-on-bind trade-off + firewall note).

The code itself is clean — `verifyConnection()` probing before declaring success (401/403 ⇒ token rejected, conn error ⇒ unreachable, 404 ⇒ OK-but-old) is the right call, and the never-trap skip/banner/re-entry flow is solid.

Resolving the conflict by taking this branch's (newer) version of both files, then merging. **Recommend tightening PR hygiene** so feature code doesn't ride in on an unrelated fix PR again.

PR #169, opened 2026-06-26
feat(mac): bind backend LAN-reachable by default, token-gated (decision ②)

oratis:
Decision ② from `docs/PLAN_IOS_ONBOARDING_v1.0.md` — the Mac-side companion to the iOS onboarding flow (#167), as a separate sub-PR.

## What
The menu-bar app starts the backend **LAN-reachable by default** (`lisa serve --web --host 0.0.0.0`) so a paired phone can reach the Mac over Wi-Fi without the user remembering `--host 0.0.0.0` — the #1 pairing failure the onboarding flow was working around.

## Why it's safe (token-gated)
Verified against `src/web/server.ts` (unchanged here):
- The server **refuses a non-loopback bind without `LISA_WEB_TOKEN`** (server.ts:185) — binding `0.0.0.0` *requires* a token; this change provides one.
- LAN requests are rejected unless they present `LISA_WEB_TOKEN` **or** a valid per-device token (568–583).
- **Loopback stays tokenless** — the local owner / island / heartbeat keep working.
- **`/api/pair/start` is loopback-only** (922–927) — a LAN peer can't mint a device token.

So `start()` mints a 24-byte random `LISA_WEB_TOKEN`, persists it to `~/.lisa/config.env` (`0600`, reused across launches, shared with `lisa pair` / a manual `lisa serve`), and injects it into the backend's environment. The phone authenticates with its own per-device token from `lisa pair`; this token just arms the gate. A `~/.lisa/serve-command.txt` override keeps full host control.

## Scope
Single file — `packaging/mac-client/Sources/Lisa/BackendController.swift`. **No server / TS changes.**

## Verification
- `swift build --package-path packaging/mac-client` → **Build complete** (only a pre-existing `WKProcessPool` deprecation warning).
- Loopback probe (`HEAD http://localhost:5757/`) is unaffected (loopback is trusted).

### DIFF
--- a/packaging/mac-client/Sources/Lisa/BackendController.swift
+++ b/packaging/mac-client/Sources/Lisa/BackendController.swift
@@ -150,12 +150,22 @@ final class BackendController {
     }
 
     /// Read KEY=value from a flat env file (first match), unquoting a simple value.
+    /// Tolerates a leading `export ` and a trailing ` # comment` the way the backend's
+    /// own parser does (src/env.ts `parseEnv`), so a hand-edited config.env doesn't
+    /// make us miss an existing token and mint a duplicate.
     private func readEnvValue(_ key: String, from path: String) -> String? {
         guard let contents = try? String(contentsOfFile: path, encoding: .utf8) else { return nil }
         for raw in contents.split(separator: "\n") {
-            let line = raw.trimmingCharacters(in: .whitespaces)
+            var line = raw.trimmingCharacters(in: .whitespaces)
+            if line.hasPrefix("export ") {
+                line = String(line.dropFirst("export ".count)).trimmingCharacters(in: .whitespaces)
+            }
             guard line.hasPrefix("\(key)=") else { continue }
-            let value = String(line.dropFirst(key.count + 1))
+            var value = String(line.dropFirst(key.count + 1)).trimmingCharacters(in: .whitespaces)
+            // Strip an inline comment, but only on an unquoted value (mirrors env.ts).
+            if !value.hasPrefix("\"") && !value.hasPrefix("'"), let hash = value.firstIndex(of: "#") {
+                value = String(value[..<hash]).trimmingCharacters(in: .whitespaces)
+            }
             return value.trimmingCharacters(in: CharacterSet(charactersIn: "\"' "))
         }
         return nil
--- a/packaging/mac-client/Sources/Lisa/BackendController.swift
+++ b/packaging/mac-client/Sources/Lisa/BackendController.swift
@@ -11,8 +11,16 @@
 //  Start command resolution:
 //    1. ~/.lisa/serve-command.txt — a full command line, if present (escape
 //       hatch for running from source, e.g. "node /path/dist/cli.js serve --web").
-//    2. otherwise `lisa serve --web` — resolved on the login shell's PATH
-//       (covers `npm i -g @oratis/lisa` / Homebrew installs).
+//    2. otherwise `lisa serve --web --host 0.0.0.0` — resolved on the login
+//       shell's PATH (covers `npm i -g @oratis/lisa` / Homebrew installs).
+//
+//  LAN-reachable by default (decision ②): a paired phone can reach the Mac over
+//  Wi-Fi without the user remembering `--host 0.0.0.0`. This is safe only because
+//  the backend is token-gated — the server REFUSES a non-loopback bind without
+//  LISA_WEB_TOKEN and rejects unauthenticated LAN requests (server.ts) — so we
+//  mint + persist a token (~/.lisa/config.env, 0600) and pass it in the backend's
+//  environment. Loopback (the local owner) stays tokenless; pairing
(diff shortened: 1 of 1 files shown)


oratis:
## Review — safe to merge; the token-gating argument holds up

I checked every server-side claim in the PR description against `src/web/server.ts` and they're all accurate:

- **`server.ts:185`** — a non-loopback `host` with no `LISA_WEB_TOKEN` throws. Binding `0.0.0.0` *requires* a token; this PR provides it. ✅
- **`server.ts:568–583`** — the auth gate rejects unauthenticated non-loopback callers (authorized by the global `LISA_WEB_TOKEN` **or** a non-revoked per-device token). ✅
- **`server.ts:922–927`** — `/api/pair/start` is loopback-only, so a LAN peer cannot mint a device token. ✅
- **`UInt8.random(in:)`** routes through Swift's `SystemRandomNumberGenerator`, which is a CSPRNG on Apple platforms (`arc4random_buf`/`getentropy`) — so the "matches `openssl rand -hex`" comment is genuinely true, not just aspirational. ✅
- **Token consistency** — `loadConfigEnv()` is *existing-env-wins* (`env.ts:80`: only sets a key when `process.env[key]` is unset/empty) and the CLI calls it at startup (`cli.ts:354`). Since `ensureWebToken()` reads `config.env` first, the injected env value and the file value are always the same token. No desync. ✅

### Design tradeoff worth a sentence in the plan (not a blocker)
Binding `0.0.0.0` by default exposes the port on **every** network the Mac joins — including untrusted Wi-Fi — for **all** menu-bar users, even those who never pair a phone. It's token-gated, so the realistic residual exposure is (a) a future auth-gate regression, (b) DoS, (c) "this Mac runs LISA" fingerprinting.

Because pairing is *loopback-only* (the QR is minted locally on the Mac), there's a strictly-least-privilege alternative that doesn't reintroduce the chicken-and-egg: **bind loopback until the first device is paired, then rebind `0.0.0.0`.** The cost is one server restart on first pair — which is presumably why decision ② just always binds. That's a defensible call, but it'd be worth one line in `PLAN_IOS_ONBOARDING_v1.0.md` on *why* always-on wins over gate-on-paired, since the security posture is the main thing a reviewer will question.

### UX gap
First `0.0.0.0` bind may trigger the macOS Application Firewall "allow incoming connections" prompt for the `node` binary. If the user dismisses/denies it, pairing fails silently — and the iOS onboarding's "Can't reach your Mac" help (`OnboardingFlow.swift` `failureView`) doesn't mention the firewall. Worth a bullet there.

### Minor / robustness
1. **Swift `readEnvValue` is a stricter parser than the TS `parseEnv` (`env.ts:86`).** It doesn't strip an `export ` prefix or inline `# comments`. A hand-edited `export LISA_WEB_TOKEN=…` line would be missed → a *second* token gets appended to `config.env`. The app-written file round-trips fine, so this only bites hand-edited configs — but a one-line comment ("only reads app-written values") or matching `parseEnv`'s leniency would avoid a confusing duplicate.
2. **The env injection is effectively redundant for the default command** — `loadConfigEnv()` already loads the token and existing-env-wins keeps them equal. It's only load-bearing for a `serve-command.txt` override that bypasses `cli.ts`. Harmless and defensive; just flagging it's belt-and-suspenders, not the mechanism.
3. `appendConfigEnvLine` writes then `chmod 0600`; `write(toFile:atomically:)` renames a temp file created at the current umask, so there's a brief window before the chmod. Single-user Mac, low severity.
4. `ensureWebToken`/`appendConfigEnvLine` aren't concurrency-guarded (two appends → two `LISA_WEB_TOKEN` lines; `readEnvValue` takes the first). `start()`'s `isStarting` flag makes a double-start unlikely, so just noting.

Nothing here blocks merge.

oratis:
### Review — token-gated LAN bind ✅

Walked the safety model against `src/web/server.ts` on `main`; every claim holds:
- Non-loopback bind **refuses** to start without `LISA_WEB_TOKEN` (`server.ts:185`).
- Non-loopback requests are rejected unless they present a token (`server.ts:568`).
- `/api/pair/start` is **loopback-only** → a LAN peer can't mint a device token (`server.ts:922-924`).
- Loopback stays trusted/tokenless (`server.ts:121`).

So minting + injecting `LISA_WEB_TOKEN` is exactly what arms the gate the `0.0.0.0` bind needs. `randomHexToken` uses `UInt8.random` → `SystemRandomNumberGenerator`, which is CSPRNG-backed on Apple platforms. Reusing an existing token from `config.env` keeps the menu-bar app, `lisa pair`, and a manual `lisa serve` consistent. 👍

**Minor (non-blocking):** `appendConfigEnvLine` writes `atomically: true` then `chmod 0600` on the final path — the temp file briefly exists at the umask default (often `0644`) before the rename, so the secret is world-readable for an instant. Low risk in `~/.lisa` on a single-user Mac, but creating the file `0600` up front would be tidier.

LGTM — merging.

PR #209, opened 2026-07-02
feat(infra): Anthropic reverse-proxy relay for Cloud Run

oratis:
Transparent Anthropic relay on Cloud Run (`packaging/gcp-relay/`) so a China-side client can reach Claude reliably by pointing `ANTHROPIC_BASE_URL` at it instead of a flaky local proxy.

- ~110-line zero-dep Node relay; forwards `/v1/*` incl. streaming to `api.anthropic.com`.
- **Key-swap gate**: real Anthropic key only in Secret Manager (injected server-side); clients present a revocable `RELAY_TOKEN` as `x-api-key`.
- **Not** an OpenClaw billing proxy — no `anthropic-beta:claude-code` spoof / OAuth reuse; bills normally (researched: those get fingerprinted + blocked by Anthropic).
- `deploy.sh` = secrets + IAM + deploy. Live on oratis-491316/us-central1.

Gotcha documented: `loadConfigEnv` doesn't override an existing env var, so a launchd/shell `ANTHROPIC_BASE_URL` shadows config.env → serve-command.txt uses `env -u`.

### DIFF
--- a/packaging/gcp-relay/Dockerfile
+++ b/packaging/gcp-relay/Dockerfile
@@ -1,0 +1,6 @@
+FROM node:20-slim
+WORKDIR /app
+COPY package.json index.mjs ./
+ENV NODE_ENV=production
+# Cloud Run injects PORT; index.mjs reads it (defaults to 8080).
+CMD ["node", "index.mjs"]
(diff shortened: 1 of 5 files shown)


oratis:
Reviewed. Security ✓: no hardcoded secrets in the diff (the `sk-ant-...` in README/comments are placeholders; `RELAY_TOKEN` is `openssl rand -hex 24`). Real Anthropic key stays in Secret Manager and is injected server-side; clients present a revocable `RELAY_TOKEN` via `x-api-key`, compared constant-time-ish after a length check. Relay only proxies `/v1/*` + health, strips client auth before swapping the real key, streams SSE straight through, zero deps. Self-contained under `packaging/gcp-relay/`, +247/-0, CI green. LGTM.

PR #233, opened 2026-07-14
fix(cli): route mail subcommand flags to its handler (--email no longer "unknown flag")

oratis:
## Problem

`lisa mail connect --email you@gmail.com` fails with:

```
unknown flag: --email
```

`mail` is dispatched as a subcommand whose args are handed **verbatim** to `runMailCommand` (`cli.ts`), but it was **missing from `RAW_SUBCOMMANDS`**. So the top-level flag parser ran first and:

- **rejected** `--email` / `--pass` / `--client-id` / `--client-secret` / `--label` as "unknown flag", and
- **silently swallowed** `--provider` / `--host` / `--port` as *global* flags, so they never reached the mail handler (e.g. `--provider gmail` set the global provider instead of selecting the Gmail OAuth flow).

Net effect: the entire `lisa mail connect` CLI was unusable in 0.16.0.

## Fix

- Add `mail` to `RAW_SUBCOMMANDS`.
- Once **any** raw-args subcommand token has appeared, collect every following token verbatim, so global-flag handlers can no longer steal a subcommand's own flags. (This also fully realizes the pre-existing "flags after a raw subcommand are command-specific" intent that `heartbeat` / `autostart` only got partially.)
- Simplify the now-dead raw branch; export `parseArgs`.

Global flags **before** the subcommand still apply (e.g. `lisa --model X mail connect …`), and an unrecognized `--flag` in global position still throws.

## Verification

```
$ lisa mail connect --email wangharp@gmail.com --pass … --host imap.gmail.com
✓ connected wangharp@gmail.com (gmail-xxxxxx). Mail consent granted.
```

- `--totallybogusflag` in global position → still `unknown flag` ✓
- `heartbeat … --flag` pass-through → unaffected ✓
- `tsc --noEmit` clean · full test suite 919/919 ✓

### DIFF
--- a/src/cli-args.test.ts
+++ b/src/cli-args.test.ts
@@ -1,0 +1,72 @@
+import { test, describe } from "node:test";
+import assert from "node:assert/strict";
+import { parseArgs } from "./cli-args.js";
+
+describe("parseArgs — raw / passthrough subcommand routing", () => {
+  test("mail: every trailing flag reaches the handler verbatim, even would-be global ones", () => {
+    const a = parseArgs([
+      "mail", "connect",
+      "--email", "me@gmail.com",
+      "--host", "imap.gmail.com",
+      "--port", "993",
+      "--provider", "gmail",
+    ]);
+    assert.equal(a.subcommand, "mail");
+    assert.deepEqual(a.subargs, [
+      "connect",
+      "--email", "me@gmail.com",
+      "--host", "imap.gmail.com",
+      "--port", "993",
+      "--provider", "gmail",
+    ]);
+    // …and none of those were consumed as global settings:
+    assert.equal(a.host, "127.0.0.1");
+    assert.equal(a.port, 5757);
+  });
+
+  test("autostart: recognized global flags are parsed into the global fields (not swallowed)", () => {
+    const a = parseArgs([
+      "autostart", "install",
+      "--port", "8080",
+      "--channels", "imessage,sms",
+      "--imessage",
+    ]);
+    assert.equal(a.subcommand, "autostart");
+    assert.deepEqual(a.subargs, ["install"]);
+    assert.equal(a.port, 8080);
+    assert.deepEqual(a.serveChannels, ["imessage", "sms"]);
+    assert.equal(a.serveImessage, true);
+  });
+
+  test("autostart: an unrecognized flag is still collected verbatim for the handler", () => {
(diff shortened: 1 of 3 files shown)


oratis:
## Review — one fix before merge ⚠️

The `mail` fix itself is correct: `lisa mail connect --email … --host … --provider …` now reaches `runMailCommand` verbatim, `lisa --model X mail connect …` still applies the global flag first, and an unknown flag in global position still throws. The value-token-vs-subcommand-token confusion I went looking for (`lisa --model mail`) is **not** present — every value-taking flag consumes its value via `mustNext(argv, ++i, …)`, so the value never re-enters the loop to trip the guard. 👍

**But the broadened guard silently regresses `autostart` and `heartbeat`.** Moving the "collect every trailing token verbatim" check to the *top* of the loop (before the `--port` / `--channels` / `--imessage` / `--model` branches) means those **recognized global** flags now get shovelled into `subargs` instead of parsed — yet the handlers still read them off the global fields:

- `autostart install` reads `args.port` / `args.serveChannels` / `args.serveImessage` (`cli.ts:430-432`)
- `heartbeat run` reads `args.model` (`cli.ts:650`)

Concrete failure:

```
lisa autostart install --port 8080 --channels imessage --imessage
```

→ before this PR: installs the LaunchAgent on 8080 with the imessage channel.
→ after this PR: `--port`/`--channels`/`--imessage` land in `subargs`, so it installs on the **default 5757 with no channels**, silently. Same shape for `lisa heartbeat run --model X` → the model is ignored.

Root cause: `mail` (wants *all* trailing flags passed through untouched) and `autostart`/`heartbeat` (want `--port`/`--channels`/`--model` parsed into the global fields) have **opposite** needs — one blanket "verbatim after any raw subcommand" rule can't serve both.

**Fix I'm pushing:** split the two behaviors —

- `PASSTHROUGH_SUBCOMMANDS = {mail}` drives the top-of-loop verbatim collection (mail truly needs it), and
- `RAW_SUBCOMMANDS = {heartbeat, autostart}` keeps the **narrow** behavior it had on `main` — only *unrecognized* flags are collected verbatim, so `--port`/`--channels`/`--imessage`/`--model` still reach the global parse.

Plus a `parseArgs` unit table (mail passthrough / autostart global flags / heartbeat `--model` / global-flag-before-subcommand / unknown-flag-throws) — there were none, and this is the exact class of change that wants one.

PR #234, opened 2026-07-14
feat(mail): guided connect modal + verify credentials before saving

oratis:
Connecting a mailbox from the Mac app used to drop the user in front of a bare
form (Email / App-password / IMAP host) with no hint of what an "app-password"
is or where to get one — and the web connect endpoint stored whatever was typed
**without verifying it**, so a wrong password produced a silently empty digest
(sweep errors are swallowed per-account). This makes both ends self-explanatory.

## Frontend — guided connect modal (`lisa-client.ts`, `lisa-css.ts`)

- A provider picker: **Gmail / iCloud / QQ / 163 / Outlook / Other** (Gmail default).
- Per-provider **numbered setup steps** + a prominent **"Open App Passwords ↗"**
  link (Gmail → `myaccount.google.com/apppasswords`). In the Mac app the link
  opens in the system browser (`target=_blank` → `NSWorkspace.open`).
- Labels / placeholders / IMAP host adapt to the provider; typing an address
  **auto-detects** the provider from its domain.
- New `.mm-providers` / `.mm-chip` / `.mm-help` / `.mm-steps` / `.mm-link` CSS.

## Backend — verify before saving (`service.ts`, `server.ts`)

- New `probeAccount()` attempts a single-message fetch (20s timeout) to confirm
  the credentials actually sign in.
- `POST /api/mail/connect` now **verifies before storing** and returns a
  plain-language error instead of persisting an unusable mailbox, e.g.:
  *"Gmail rejected the sign-in. Use a 16-character app password (not your Google
  login password), and make sure 2-Step Verification is on."*
  (`friendlyMailError()` maps auth vs host vs network/timeout.)

Still read-only; still IMAP + app-password (Gmail via `imap.gmail.com`). OAuth
in the web UI and persistent sweep-error surfacing are follow-ups.

## Verification

Driven in a browser against the real served `MAIN_HTML`:

- Gmail default renders steps + link + `imap.gmail.com`; switching to QQ swaps to
  授权码 steps / `imap.qq.com`; typing `me@gmail.com` snaps back to Gmail.
- Submitting a wrong password shows the plain-language error inline (modal stays
  open to retry) — no console errors.
- `MAIN_HTML` snapshot refreshed; `tsc --noEmit` clean; full suite 919/919.

Note: `src/cli.ts` carried a separate `lisa mail connect` parser bug — fixed in
its own PR (#233), kept out of this one.


oratis:
## Review — one fix before merge ⚠️

Design is clean and "verify before store" is the right call. Security checks out: the modal's dynamic parts (steps/chips/link) are all `createElement` + `textContent` + `setAttribute`, the typed email/password only ever go into the `fetch` body (never the DOM), and the app-passwords link is `https` + `target=_blank` + `rel=noopener`. `POST /api/mail/connect` correctly refuses to persist when the probe fails, with no double-`res.end` / partial write. The snapshot delta (+9649 B) is confined to this PR's own provider JS/CSS/HTML.

**MUST-FIX — `probeAccount` leaks a live IMAP session on the timeout race** (`src/mail/service.ts:32-50` × `imap.ts` `close()`):

- `close()` runs in the outer `finally`, which fires the moment `Promise.race([run, timeout])` settles — i.e. as soon as the 20s timeout wins.
- At that instant `client.connect()` is still in flight, so `this.connected` is still `false`, and `close()` short-circuits on `if (!this.connected) return` — a **no-op**.
- The still-running `listSince` then finishes connecting (`connected = true`), opens INBOX, fetches, and returns — leaving a **fully-established IMAP session that nobody ever closes.** Every retry against a slow host stacks another leaked socket.

This is exactly the case the 20s timeout exists for: a host slower than 20s but ultimately reachable. `run.catch(() => {})` only silences the rejection path; it doesn't clean up the *success* side-effect (the open connection).

**Fix I'm pushing:** chain the close onto `run` **settling** (not just the race) —

```js
} finally {
  // Close once the underlying op actually settles, not merely when the race
  // does: if the timeout won, connect() may still be in flight, so closing now
  // would no-op (connected===false) and leak the session it later establishes.
  void run.then(() => connector.close().catch(() => {}),
               () => connector.close().catch(() => {}));
}
```

By the time `run` resolves, `connected` is `true`, so `close()` really logs out. I'm also filling the couple of `inferHost` gaps the new modal advertises but the backend doesn't map (`mac.com` → `imap.mail.me.com`, `foxmail.com` → `imap.qq.com`, `msn.com` → `outlook.office365.com`, `yeah.net` → `imap.yeah.net`), so a left-blank host matches the placeholder the UI shows instead of failing the probe.

PR #238, opened 2026-07-14
feat(web): render Lisa's Markdown output as styled HTML

oratis:
## What

Lisa emits standard Markdown, but the web client rendered her replies with `textContent` — so `###` / `**` / `-` / ``` / tables showed up as literal symbols. This adds a renderer and wires it into every surface that shows her prose.

## Changes
- **`md-render.ts` — `renderMarkdown()`**: headings, bold/italic, inline + fenced code (lang label + copy), nested ordered/unordered lists, blockquotes, `<hr>`, tables, and links restricted to `http`/`https`/`mailto`. **Escape-first**: untrusted model output is HTML-escaped before any tag is introduced — no `javascript:`/`data:` hrefs, no raw HTML passthrough.
- **Source-injected** into the page like `agent-roster.ts → island.ts`, so the browser runs the exact, unit-tested function instead of a hand-escaped copy inside the no-interpolation client literal. A `__name` shim absorbs esbuild `keepNames` wrappers when the server runs under tsx.
- **`lisa-client.ts`**: the streaming bubble accumulates raw text and re-renders once per animation frame (flushed at segment/turn boundaries); history bubbles and idle "★" cards use the same function; code-block copy via event delegation. The user's own bubbles stay plain text.
- **`lisa-css.ts` / `MD_RENDER_CSS`**: styled elements themed to the existing dark tokens; chat + idle cards share one block (`:is(.msg, .idle-block)`), while the standalone Room/Island pages share `MD_RENDER_CSS`.
- **`room.ts` / `island.ts`**: Lisa's ★ reflection notes render Markdown too (Room joins piled-up notes with an `<hr>`); UI-phrase fields stay plain.

## Tests
- New `md-render.test.ts` (19): structure, XSS + link-scheme safety, source-injection stands alone.
- `html-syntax.test.ts` now also compiles `ROOM_HTML` (all three inline scripts parse).
- `MAIN_HTML` byte snapshot refreshed.
- Verified in a browser with the real CSS + renderer (chat, idle, Room/Island).

_Note: a pre-existing `imapflow`-missing failure in `auth.test.ts` is unrelated to this change (it fails on `main` too)._


oratis:
## Review — critical fix before merge 🔴

The escape-first XSS architecture holds — I tried hard to break it (mixed-case `JaVaScRiPt:`, `java\tscript:`, `&#106;avascript:`, attribute-breakout via `"`, `data:text/html`, code-fence lang injection, ReDoS on 200k-char single lines) and every path renders inert. `esc()` is applied to each raw run *before* any tag is introduced, the scheme allowlist survives the entity/whitespace tricks, and the copy button reads `pre.textContent` (no injected payload). Streaming flush is correct too — `tool_start`/`done` both paint synchronously, so the final segment is never lost. The safety model is genuinely well built. 👍

**But there's a release-blocking infinite loop.** The fence **opener** `/^\s*```(\w*)\s*$/` is stricter than `isFence` (`/^\s*```/`). A fence whose info string contains any non-word char matches **neither** the opener **nor** the paragraph gatherer's `!isFence(...)` guard — so the paragraph `while` body never runs, `i` is never incremented, and `renderMarkdown` spins forever appending `<p></p>` until the tab OOMs.

It triggers on completely ordinary output — <code>```c#</code>, <code>```c++</code>, <code>```objective-c</code>, <code>```js title="app.js"</code>, or any lang containing `-` `#` `+` `.` or a space. And because the renderer runs **inside the streaming `requestAnimationFrame` paint**, the tab freezes mid-stream the instant that token arrives — no adversary required, normal Lisa answers do it. It also hangs the Room/Island/history render on any stored note containing such a fence. Minimal repro:

<pre>
```c#
var x = 1;
```
</pre>

The 19 tests are all happy-path fences, so they miss it entirely.

**Fix I'm pushing:** match the opener to any info string and keep the first token as the (escaped) lang — `/^\s*```(\S*)/` — so `isFence` and the opener agree; a forward-progress guard in the paragraph branch as belt-and-suspenders; and a regression test for <code>```c#</code>. I'm also tokenizing links out before the emphasis passes (right now a `**`/`` ` `` inside a URL gets transformed inside the `href` — inert, but it produces dead links, e.g. `[x](https://e.com/**b**)`).

PR #242, opened 2026-07-14
fix(web): reflect on idle so web conversations evolve Lisa's desires

oratis:
> **Stacked series (1/3).** This is the base of a 3-PR stack fixing why desires don't evolve. Order: **#242 → #246 → #247**. #246 and #247 are branched on top of this one.

## Why

Investigating *"the wish never changes no matter what we talk about"* traced to a dead code path: **`reflectOnSession` — the only automatic way a conversation turns into or updates a desire — is never triggered in the web deployment.** It's wired to CLI `--reflect` and the channel router only; the `POST /reflect` endpoint exists but **no client calls it**. So web conversations never touched Lisa's desires. This is the dominant root cause (full analysis in the plan doc).

## What

A dedicated, debounced **reflection scheduler** in `startWebServer`:

- After `LISA_REFLECT_DEBOUNCE_MS` (**default 5 min**) of no user input, reflect **once** on the accumulated history.
- Only when the **user added a new message** since the last reflection — counting *user* messages (not total) so Lisa's own idle `[while you were away]` notes never trigger reflection on her own monologue.
- **Independent** of the 60-min "dream" idle (`src/cli.ts` default) — works even with `--no-idle`. It reuses the process-wide idle watcher purely as an activity clock (`/chat` already `tick()`s it).
- Robust: advances its marker only on success (a failed reflect retries next tick); snapshots history before the async call; `unref()`s its timer; skips while a reflection or idle run is in flight.
- Emits an SSE `reflect_done` event and appends to the session log for observability.

Decision logic is extracted to a pure, unit-tested module (`src/web/reflect-scheduler.ts`).

## Plan doc

This is **PR 1 of 3** from [`docs/PLAN_DESIRE_EVOLUTION_v1.0.md`](docs/PLAN_DESIRE_EVOLUTION_v1.0.md), which bundles the root-cause analysis, the pro/con debates (正反方辩论), and the PR1–3 breakdown. PR2 makes reflection able to *revise/close* desires (not just append); PR3 surfaces the most recently *active* desire instead of a filesystem-order pick.

## Testing

- `src/web/reflect-scheduler.test.ts` — 8 tests covering the debounce/guard truth table and user-message counting. All green.
- `npm run typecheck`: no new errors (pre-existing `imapflow` optional-dep error only).
- `npm test`: **zero regressions** — same 5 pre-existing failures as `main` (imapflow mail tests, a stale HTML snapshot, an auth test), plus the 8 new passing tests.

Backend timer change — not observable in the browser preview; verified via unit tests + typecheck.

### DIFF
--- a/docs/PLAN_DESIRE_EVOLUTION_v1.0.md
+++ b/docs/PLAN_DESIRE_EVOLUTION_v1.0.md
@@ -1,0 +1,248 @@
+# PLAN — Desire Evolution v1.0
+
+> Why Lisa's "wishes" (愿望 / desires) don't change as a conversation evolves, and
+> the plan to fix it without breaking the soul's deliberate, sovereign design.
+
+Status: **in progress**. Owner: Lisa + oratis. Created 2026-07-15.
+
+---
+
+## 0. TL;DR
+
+The observed complaint — *"the wish never changes no matter what we talk about"* — is
+**three-quarters bug, one-quarter by-design**. Concretely:
+
+1. **The conversation→desire path never fires in the web deployment.** `reflectOnSession`
+   is the only automatic way a conversation turns into (or updates) a desire, and it is
+   never triggered from the web server. The `POST /reflect` endpoint exists but no client
+   calls it. This is the dominant cause and a genuine bug.
+2. **Reflection is append-only and blind.** Even when it runs (CLI `--reflect`, channels),
+   it can only `desire_add`; it is never shown the existing desires, so it cannot revise,
+   dedupe, or prune. Desires pile up and never evolve.
+3. **The displayed "current desire" is a fixed pick by filesystem order**, decoupled from
+   both recency and the live conversation: `desires.find(d => d.actionable)` over
+   `fs.readdir` order.
+4. **By design (keep this):** desires are Lisa's identity, meant to evolve *slowly and
+   deliberately* through reflect / heartbeat / weekly_examen — not thrash on every message.
+   The fix must make desires **responsive, not reactive.**
+
+This plan ships **three focused PRs** plus this doc, and explicitly **defers** a fourth
+(intra-session live "focus") pending results. Every decision below is accompanied by a
+pro/con debate (正反方辩论) and a recorded decision.
+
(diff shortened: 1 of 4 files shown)


oratis:
## Review — merge-ready ✅

Walked the concurrency + debounce truth table against the code:

- **`--no-idle` really works** (the central claim) — `/chat` calls `getIdleWatcher(...).tick()` unconditionally, so the scheduler's activity clock advances even when the 60-min idle block is skipped. The scheduler only uses `idleFor()` + its own interval; it never needs the watcher's `idle` event.
- History `snapshot` + `snapshotUserCount` + `reflecting = true` are set in one synchronous block (no `await` between), and the marker advances to the snapshot count **only on success** — so a message arriving mid-reflect is counted exactly once next tick, never lost or double-counted.
- `reflecting` is set synchronously before the async IIFE and reset in `finally` (even on a sync throw), so two reflects can't overlap; a failed reflect retries without advancing the marker.
- Soul writes serialize on the cross-process file locks and session-log appends are `O_APPEND` atomic, so a reflect + the dream can't corrupt shared state.

Sound — no blockers. Two small optional hardenings I'll fold in while merging:

1. **Clamp `LISA_REFLECT_DEBOUNCE_MS` to `> 0`.** `Number(env) || DEFAULT` correctly rejects `NaN`/`""`/`"0"`, but a negative value is truthy → `debounceMs = -1` makes `idleMs < debounceMs` always false, so it reflects ~60s into any conversation. `Number.isFinite(n) && n > 0 ? n : DEFAULT`.
2. **Have the 60-min dream also respect `reflecting`.** The scheduler's `inFlight: reflecting || idleRunning` blocks reflect-during-dream, but the dream's guard is only `if (idleRunning) return` — so under a short `--idle` (both firing on the same quiet trigger) a dream can start mid-reflect and its own `[while you were away]` note can bleed into the reflected history — the exact thing PLAN §3 wants to avoid. Adding `|| reflecting` to the dream guard closes the asymmetry.

Neither blocks the fix; merging after those.

PR #245, opened 2026-07-14
feat(web): 九宫格 nav (3×3) with Mail + Settings rail views

oratis:
## What & why

The left-rail nav was a 7-item vertical list. This reworks it into a **3×3 九宫格 tile grid**, growing to **9 tiles** by adding two real destination views — **Mail** and **Settings** — to fill the last two slots.

## Changes

**Nav → 九宫格**
- 9 icon tiles in a `repeat(3,1fr)` grid with a **unified line-SVG icon set** (matching the top function bar); active-tile highlight; corner count badges for Control (agents) and **Mail (needs-you)**.
- Tiles use `min-height` instead of `aspect-ratio` — fixes a blow-up where tiles became ~480px tall in compact/full-width (stacked) mode.

**Mail view** (`#viewMail` / `loadMail`)
- Reuses `/api/mail/*`: digest summary, needs-you list, sweep-now, connect-mailbox modal.
- Adds **per-account enable / disable / remove** (the sidebar card lacked this) via the existing `/api/mail/accounts/:id/:action` route.

**Settings view** (`#viewSettings` / `loadSettings`)
- **API keys** — status + update form (`/api/config/status|save`, loopback-only save already enforced server-side).
- **Proactive** autonomy toggle (`/api/autonomy/state`) and **Compact** mode toggle — both **relocated out of the sidebar footer** into Settings (single home; kept in sync via `setProactiveUI` and shared `lisaSetCompact`/`lisaGetCompact`).
- **About** — edition from `/api/edition`.

**Snapshot**
- Regenerated the byte-exact `MAIN_HTML` length + sha256 constants and appended a changelog note.

## Verification
- Drove the UI in a browser (throwaway harness serving the real `MAIN_HTML` + canned JSON): all 9 tiles switch views; grid + SVG icons render; Control/Mail badges show; Mail digest/accounts/sweep work; Settings key-status, Proactive round-trip, and Compact stacked layout all work; **no console errors**.
- `lisa-html-snapshot.test.ts` ✅ passes; `typecheck` clean except a **pre-existing, unrelated** `imapflow`-not-installed error in `src/mail/connectors/imap.ts` (verified identical with my changes stashed). This diff only touches `src/web/lisa-*`.

## Note
The `imapflow` test/typecheck failures are environmental (declared dep not installed in this checkout) — run `npm install` to green them; not addressed here to avoid unrelated `node_modules` churn.


oratis:
## Review — clean code; one decision to make at merge ⚠️

Reviewed the PR's own diff in context and it's solid: all 9 tiles map to real views (`mail`→`loadMail`, `settings`→`loadSettings` added to `views` + the `loadView` chain; `showView` still falls back to `chat`), every endpoint the new Mail/Settings views call exists (`/api/mail/accounts/:id/{enable|disable|remove}`, `/api/config/{status,save}` — loopback-gated + validated, `/api/edition`), no XSS (all model/user text goes through `esc()` into double-quoted attributes; API keys are never echoed back), and the CSS is self-consistent (300px sidebar fits 3 columns, overflow handled). The Mail rail view **reuses** #234's connect modal + endpoints rather than reimplementing them, and `refreshMail` repaints both surfaces — nicely coordinated.

**The one thing to get right on merge:** this branch's diff was written against a **7-item** nav, but v0.17.0 shipped an **8th** view — `kb` / "Knowledge" (#240–243). The PR hardcodes exactly **9** tiles (chat, dashboard, control, reve, room, sense, memory, mail, settings) with no `kb`, so a naïve merge silently **drops Knowledge from the nav** and orphans `#viewKb` / `loadKb` / `loadKbList` / the `loadView` `kb` branch — 10 views competing for 9 九宫格 slots. That's a design call, not a textual conflict.

I'll resolve it deliberately so Knowledge stays reachable (rather than letting it drop silently), and recompute the snapshot constants after the v0.17.0 merge.

Non-blocking: two mail surfaces now coexist (the #234 sidebar card + this full Mail view) — fully coordinated, but you may want to retire the sidebar card later.

PR #248, opened 2026-07-15
feat(web): surface the desire the conversation is about (intra-session focus)

oratis:
> **Stacked PR (4/4).** Base is `claude/desire-current-pick` (#247) — GitHub retargets automatically as parents merge. Merge order: **#242 → #249 → #247 → this**. Diff below is only this PR's own changes.

## Why

PR1–3 make desires change at reflection **pauses**. This closes the last gap from Debate 4: making the *surfaced* desire track the conversation **turn-by-turn** — the finest-grained version of "the wish follows what we talk about." Originally deferred (§6 F1); implemented now at the owner's request ("全部进行"), staying strictly inside the guardrails I set for it.

## What

- Pure `pickFocusedDesire(desires, recentText)` + `tokenize` + `recentUserText` in `src/soul/desire-focus.ts`: lexical overlap between the recent user turns and each desire's `what`/`why`. **No per-turn LLM call. No persisted state. Display-only.**
- **Cross-lingual**: latin word-tokens *and* CJK character bigrams, so it works when Lisa and the user converse in Chinese (no segmenter needed).
- Wired into `GET /api/island/ping`: when the conversation is fresh (`idleFor() < FOCUS_FRESHNESS_MS`, 15 min) and *clearly* about one desire, surface it; otherwise fall back to PR3's `pickCurrentDesire`. A tie or weak match (< 2 shared tokens) returns `null`, so it **can never invent focus** — worst case it degrades to the recency pick.

## Why it's safe (Debate 4's 反方 concerns, addressed)

| Concern | Mitigation |
|---|---|
| Per-turn cost | Pure string overlap, no model call |
| Thrashing | Freshness gate + strict-winner + `FOCUS_MIN_OVERLAP=2`; ties → fallback |
| Two notions of "want" | Focus only *chooses which existing desire to show* — it never creates one |

## Testing

- `src/soul/desire-focus.test.ts` — 10 tests: tokenizer (latin + CJK bigrams), match / no-match / weak / tie, **Chinese cross-lingual match**, `recentUserText` gathering. All green.
- `npm test`: **zero regressions** — same 3 pre-existing failures as `main` (`imapflow` mail tests, `web/auth`); +10 new passing. Typecheck clean.

Backend/data-selection change; the pure logic is fully unit-tested (browser preview would need a seeded born-soul fixture).

### DIFF
--- a/docs/PLAN_DESIRE_EVOLUTION_v1.0.md
+++ b/docs/PLAN_DESIRE_EVOLUTION_v1.0.md
@@ -138,9 +138,22 @@ Closes G3.
 Files: `src/soul/store.ts`, `src/web/server.ts`, `src/cli/status.ts`,
 test `src/soul/pick-desire.test.ts`.
 
-**Recommended merge order: PR1 → PR2 → PR3.** They branch independently off `main` and touch
-disjoint regions of the two shared files (`web/server.ts`: idle handler vs ping endpoint;
-`store.ts`: revise/close vs pick helper), so conflicts are trivial-to-none.
+### PR 4 — `feat(web): surface the desire the conversation is about (intra-session focus)`
+Implements F1 (Debate 4), added at the owner's request after PR1–3. Conservative by construction.
+
+- Pure `pickFocusedDesire(desires, recentText)` + `tokenize` + `recentUserText` in
+  `src/soul/desire-focus.ts`. Lexical overlap only — **no per-turn LLM call, no persisted state.**
+- Cross-lingual: latin word-tokens **and** CJK character bigrams, so it works when Lisa and the
+  user converse in Chinese (no segmenter needed).
+- Wired into `GET /api/island/ping`: when the conversation is fresh (`idleFor() <
+  FOCUS_FRESHNESS_MS`) and *clearly* about one desire, surface it; otherwise fall back to PR3's
+  `pickCurrentDesire`. A tie or weak match returns null, so it can never invent focus.
+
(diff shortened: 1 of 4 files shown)


oratis:
## Review — SHIP-WITH-NITS ✅ (+ review fix applied)

Core module is well-built: token-set overlap scoring, tie-break returns null (traced `[3,2,3]`/`[3,3,5]`/`[5,2,5]`/`[4,3,3,4]` — the `tied` flag resets on each new strict max), empty text/desires/history all return null cleanly, `recentUserText` correctly skips assistant idle-injections, CJK-bigram path works.

**Applied on this branch (review fix):**
- `pickFocusedDesire` now filters to the **actionable** pool (mirrors `pickCurrentDesire`) so a lexically-matching **closed/dormant** desire can't be surfaced as the conversation's focus. [MED → fixed]

**Deferred follow-up (flagged, not blocking):**
- **[MED] freshness gate defeated after a server restart** (`server.ts` `reflectClock.idleFor() < FOCUS_FRESHNESS_MS`) — `idleFor()` counts from process start, but `history` is restored from disk (the backend is a launchd LaunchAgent that restarts), so a stale overnight chat can pin a stale focus for ≤15 min. Correct fix needs a per-message/transcript timestamp (bare `StoredMessage` has none) rather than the process clock — worth a focused follow-up.
- Low: CJK 2-char threshold slightly more permissive than Latin; non-Latin/non-CJK scripts unsupported (graceful); `desireActivity` computed even when `focused` short-circuits it (minor poll I/O).

Note: this branch was rebased onto current main (the stacked base #249/#250 landed first) + the actionable fix; verified locally — typecheck clean, full suite green (1010 pass).

PR #259, opened 2026-07-21
fix(review): App Store 2.1 remediation + accounts/billing plan (B0)

oratis:
## Why

The 2026-07-13 rejection (Guideline 2.1, submission `cb80235c`) said the reviewer was **unable to sign in when entering the code**. Investigation found three stacked causes — all addressed here, so this can be resubmitted without waiting for the full accounts work:

1. **Settings' Cloud "Connect" was parse-only fake success** — it never touched the network, so a bad/truncated token still showed "Connected to LISA Cloud." and the real 401 only surfaced in Chat.
2. **ASC's user/pass fields don't fit our paste-a-URL sign-in** — the reviewer had no way to enter the "code" correctly.
3. **Bare-401 + exact-match token gate** — no trim on pasted tokens, no diagnosable error.

## What

- **iOS**: Cloud Connect / Mac Apply-pairing / QR scan now run `verifyConnection` and report the real outcome (401 → actionable message; unreachable → hint). Spinner + disabled state while probing.
- **Server**: cookie/query tokens are trimmed; 401 body is structured JSON (`token_missing` / `token_mismatch` / `server_not_configured`).
- **RELEASE.md**: App Review notes rewritten — explicit "no username/password form" callout, exact paste steps, ASC field guidance (`see-review-notes` / token), pre-submission `curl` check that reproduces the reviewer's path.
- **deploy.sh**: Cloud Run sizing for the account era (2 vCPU / 4 GiB, `--timeout 3600` so SSE streams aren't cut). min=max=1 stays until B2.
- **docs**: `PLAN_ACCOUNTS_BILLING_v1.0.md` (design v1.1, decisions locked) — centralized accounts as the primary flow on all surfaces, 12h $5 session quota + paid tiers, consumable IAP, inference gateway for key-free local Macs. Follow-up PRs (B1–B7) implement it.

## Verification

- `npm run typecheck` clean; `npm test` 1018 pass / 0 fail.
- iOS simulator build (`build.sh`): **BUILD SUCCEEDED**.

## Operator actions (not in this PR)

- Re-run `deploy/deploy.sh` to apply the new sizing.
- Before resubmitting: run the curl check from RELEASE.md against the live demo URL+token, update the ASC review notes + fields per the new template, and reply to the rejection in ASC.

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -87,6 +87,9 @@ IGN
 cleanup() { rm -f ./Dockerfile ./.gcloudignore; }
 trap cleanup EXIT
 
+# Sizing (PLAN_ACCOUNTS_BILLING §6.7): 2 vCPU / 4Gi headroom for the account era;
+# --timeout 3600 so SSE chat streams aren't cut at 5 min. min=max=1 stays until
+# the per-uid + Firestore work (B2) removes the single-writer constraint.
 echo "→ deploying $SERVICE to $PROJECT/$REGION (model ${LISA_MODEL:-claude-default}, /data←gs://$BUCKET, min=max=1 single-writer, allow-unauthenticated; the app's token gate is the auth)"
 gcloud run deploy "$SERVICE" \
   --source . --project "$PROJECT" --region "$REGION" --quiet \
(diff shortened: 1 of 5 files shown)


oratis:
### 🤖 Automated review — ✅ SAFE TO MERGE

App Store 2.1 remediation (B0). Verified:
- Removing the `webToken ? … : null` guard on `presented` is **not** an auth bypass: `isRequestAuthorized` still requires `webToken && presented && timingSafeEqual` for the shared-token path, and the newly-reachable `verifyDeviceToken` path requires a stored SHA-256 hash matched in constant time. When `LISA_WEB_TOKEN` is set (the demo), behavior is unchanged.
- Structured 401 never echoes the presented token (category only). `SettingsView` now runs a real `verifyConnection()` — the actual "couldn't sign in with the code" fix. No secrets committed (placeholders).

Nits (non-blocking):
- 401 `reason` (`token_mismatch` vs `token_missing`) is a mild oracle — negligible against a high-entropy `LISA_WEB_TOKEN`.
- `webToken` isn't trimmed at load; an env token with stray whitespace becomes unmatchable (fails closed, but a config footgun) — trim at load for symmetry with the presented-token path.

PR #260, opened 2026-07-21
feat(accounts): B1 server core — accounts, sessions, auth endpoints, in-app deletion

oratis:
**Stacked on #259** (B0). Implements the server half of milestone **B1** from `docs/PLAN_ACCOUNTS_BILLING_v1.0.md`.

## What

- **`src/web/accounts.ts`** — one account store (`$LISA_HOME/accounts.json`, 0600) for both kinds:
  - **Apple** (`apple-<sub>`): upserted on every verified Sign in with Apple; no password material.
  - **Email** (`em-<random>`): scrypt-hashed passwords, normalized emails, constant-cost verify for unknown emails (no user enumeration), 5 fails → 15-min lockout.
  - `sessionVersion` per record → stateless account-wide revocation.
- **`src/web/sessions-auth.ts`** — `s1.<payload>.<hmac>` sessions ({uid, iat, exp, sv}), 30-day TTL, sliding cookie renewal past half-life. Signing secret auto-created (0600) under `$LISA_HOME` → durable on the cloud `/data` mount, **no new deploy env needed**.
- **`server.ts`**:
  - `POST /api/auth/register` / `login` / `logout` (cloud-only, pre-gate — they mint access, like `/api/auth/apple`). Typed error codes (`email_taken`, `weak_password`, `throttled`→429…).
  - `/api/auth/apple` now returns a **per-uid session** instead of the shared `LISA_WEB_TOKEN` — the uid keys per-user isolation (B2) and billing (B3+).
  - Auth gate accepts sessions (sv checked against the live record); shared-token and device-token paths untouched (BYO escape hatch preserved).
  - `GET /api/auth/me`; `DELETE /api/account` (**App Store 5.1.1(v)**) — session-authed only, removes record + `users/<uid>` subtree, expires the cookie.

## Tests

14 new tests (sessions: round-trip / tamper / expiry / renewal / secret perms; accounts: register / login / throttle / delete / no-plaintext-on-disk / store perms). Full suite: **1032 pass / 0 fail**.

## Notes

- iOS/web clients pick this up in the next PR (B1-clients): SIWA UI un-gated, onboarding flip, web login page.
- Pre-B2 all uids still share the single demo soul; isolation is the next milestone and deletion's home-cleanup is already per-uid-shaped.

### DIFF
--- a/src/web/accounts.test.ts
+++ b/src/web/accounts.test.ts
@@ -1,0 +1,99 @@
+import { test, describe, beforeEach } from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
+const TMP = fs.mkdtempSync(path.join(os.tmpdir(), "lisa-accounts-"));
+process.env.LISA_HOME = TMP;
+const FILE = path.join(TMP, "accounts.json");
+
+const {
+  createEmailAccount,
+  verifyEmailLogin,
+  upsertAppleAccount,
+  deleteAccount,
+  getAccount,
+  getAccountByEmail,
+  sessionAccountValid,
+  resetLoginThrottles,
+  appleUid,
+  AccountError,
+} = await import("./accounts.js");
+
+beforeEach(() => {
+  fs.rmSync(FILE, { force: true });
+  resetLoginThrottles();
+});
+
+describe("email accounts", () => {
+  test("register → login round-trip; email is normalized", () => {
+    const rec = createEmailAccount("  Reviewer@MeetLisa.AI ", "correct-horse-1");
+    assert.match(rec.uid, /^em-[0-9a-f]{18}$/);
+    assert.equal(rec.email, "reviewer@meetlisa.ai");
+    assert.equal(rec.verified, false);
+    const back = verifyEmailLogin("reviewer@meetlisa.ai", "correct-horse-1");
+    assert.equal(back?.uid, rec.uid);
+  });
+
+  test("wrong password / unknown email → null (no throw, no enumeration)", () => {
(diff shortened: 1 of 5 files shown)


oratis:
### 🤖 Automated review — ✅ MERGE (nits) — auth core verified sound

Read the crypto/authz paths directly. **No auth-bypass, secret-leak, or unauthorized-delete found.**
- Sessions: HMAC-SHA256 over `s1.<payload>`, 384-bit random secret (0600), `timingSafeEqual` after length check, fixed format (no JWT `alg`-confusion), `exp` enforced, `sv` re-checked against the live record so **deleted accounts lose access immediately** (`accounts.ts:399`).
- Passwords: scrypt(N=16384) + 16-byte salt, constant-time compare, constant-cost **decoy** for unknown emails (no login enumeration). `appleUid` sanitizes `sub` to one path segment.
- `DELETE /api/account` is session-authed only (demo token → `accountUid=null` → 403), deletes only the caller's own uid (no IDOR), `fs.rm` guarded by `startsWith(users/)`.

Hardening (tracked for the follow-up hardening commit / backlog):
- `accounts.ts:108,119` — `scryptSync` on the **unauthenticated** `/register`+`/login` path blocks the event loop; `/register` has no rate-limit and `readJsonBody` is unbounded → CPU/memory DoS. Use async `crypto.scrypt` + per-IP cap + body-size limit.
- `server.ts` — session `Set-Cookie` omits **`Secure`** on the HTTPS cloud edition (HttpOnly + SameSite=Strict present). **Being added in the hardening commit.**
- `register` 409 `email_taken` is a user-enumeration oracle; `sessionAccountValid` re-reads+parses `accounts.json` on every authed request — add an in-memory cache invalidated on write.

PR #261, opened 2026-07-21
feat(accounts): B1 clients — SIWA on, sign-in as primary iOS flow, web login page

oratis:
**Stacked on #260** (B1 server). Client half of milestone B1: the decision that **centralized accounts are the primary flow on every surface**, with the BYO-key / Mac-pairing path preserved as the always-visible escape hatch.

## iOS

- **Sign in with Apple is ON**: `applesignin` entitlement restored in `project.yml` + `LISA_ENABLE_SIWA` compiled in. The original reason for gating it off (5.1.1(v) account deletion) is resolved — deletion ships in the same build.
- **Shared `CloudSignInForm`** (new `AccountViews.swift`): SIWA → email+password → "Advanced: token link" disclosure. Used by **both** Settings and onboarding so the two entry points can't drift; every path live-verifies before declaring success (the 2.1 fix from B0, now on all flows).
- **`AccountCard`**: signed-in state, Sign out, confirm-gated **Delete account** (5.1.1(v)).
- **Onboarding flipped**: LISA Cloud ("Sign in and go", *Recommended*) first; My Mac ("Advanced: private and local") second — unchanged and never hidden.
- `AppState`: `connectCloudWithEmail` / `refreshAccount` / `signOutCloud` / `deleteCloudAccount`; `defaultCloudBase` prefill (**cloud.meetlisa.ai**).
- Privacy manifest + store metadata updated: **Email Address + User ID** (app functionality, linked, no tracking).

## Web

- `src/web/login.ts`: email+password login/register page, served with **401 status** to unauthenticated browser GETs on the cloud edition (API callers keep the structured JSON 401 from B0). SIWA-on-web needs an Apple Services ID → deferred to B7.

## Verification

Typecheck + full suite green; iOS simulator build **SUCCEEDED** (with SIWA compiled in).

## Operator actions

1. **Apple Developer portal**: enable *Sign in with Apple* on the App ID `ai.meetlisa.main` (required before the next signed/TestFlight build — automatic signing will fail without it).
2. **DNS/Cloud Run**: map `cloud.meetlisa.ai` to the `lisa-cloud` service (the app prefills that URL), or change `AppState.defaultCloudBase` before release.
3. **Cloud env**: set `LISA_CLOUD_APPLE_SIGNIN=1` on deploy (deploy.sh already passes it through).
4. ASC **App Privacy**: switch answers to Email Address + User ID per `APPSTORE_METADATA.md`.

### DIFF
--- a/packaging/ios-companion/APPSTORE_METADATA.md
+++ b/packaging/ios-companion/APPSTORE_METADATA.md
@@ -75,15 +75,15 @@ Live Activities and the Dynamic Island keep the current agent's progress on your
 lock screen. Home Screen widgets show it without opening the app.
 
 PRIVATE BY DESIGN
-Lisa Pocket is a client, not a service. Pairing is QR-based and every connection
-is gated by a per-device token you can revoke. Nothing is collected to a
-Lisa-operated server; the app stores no account.
+Your choice of two ways to run: sign in to LISA Cloud (email or Apple — a free
+usage allowance refreshes every 12 hours), or connect your own Mac, where
+pairing is QR-based, every connection is gated by a revocable per-device token,
+and nothing ever reaches a Lisa-operated server.
 
 GETTING STARTED
-You'll need LISA running on a Mac (free and open source — install with Homebrew,
-npm, or the Mac app). The in-app setup walks you through installing it, starting
-it reachably, and scanning the pairing QR. No Mac yet? Point the app at a LISA
-Cloud URL instead.
+Sign in and go — no setup needed. Prefer fully local? LISA is free and open
+source: the in-app setup walks you through installing it on your Mac, starting
+it reachably, and scanning the pairing QR.
 
 LISA is open source (MIT). Learn more at meetlisa.ai.
 ```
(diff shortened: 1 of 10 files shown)


oratis:

- Session token stored in **Keychain** (`kSecClassGenericPassword`, `AccessibleAfterFirstUnlock`), not UserDefaults.
- `login.ts` is a static template (no server-side interpolation); JS errors render via `textContent` off a fixed `MSG` map → no reflected XSS.
- `verifyThenReport()` gates "success" on a real authed probe — consistent across Settings + onboarding.

Nits (non-blocking):
- **SIWA has no nonce end-to-end** (`AccountViews.onRequest` never sets `req.nonce`; server verifier doesn't check one). Replay window is bounded by `aud`=bundle-id + short `exp` + TLS, so low practical risk — but Apple-recommended; add before the multi-tenant launch.
- `AppState.parseCloudBase` (:211) accepts `http://` → email+password POSTed in cleartext if a user types an http URL. Force/warn https for the hosted base.
- The Apple handler discards Apple's `email_verified`; the record is marked `verified:true` while the email may be Apple-unverified — matters once email becomes a billing/recovery key.

PR #262, opened 2026-07-21
feat(cloud): B2 — per-uid home isolation (AsyncLocalStorage seam) + lazy per-user birth

oratis:
**Stacked on #261.** The multi-tenant core — the gate before any real cloud user beyond a reviewer.

## Design

Two homes, deliberately:
- **`lisaGlobalHome()`** — operator/process home: `config.env`, `accounts.json`, device tokens, session secret. Never per-uid.
- **`lisaHome()`** — the ACTIVE home for identity-bearing state (soul, sessions, memory, reflections, kb, autonomy…). AsyncLocalStorage-scoped: a signed-in cloud request enters `homeScope` with `$LISA_HOME/users/<uid>` and every downstream path — across awaits — resolves there.

Path **constants became functions** (`SESSIONS_DIR` → `sessionsDir()` etc., 65 files) because an import-time constant freezes the home — exactly what made the codebase single-tenant. Module-scope constants that captured the home at import were functionized too (active-session pointer, autonomy dir, embeddings cache, advisor state…).

## Server

- Per-uid **ChatCtx** `{session, history, chain}`: each account gets its own conversation, restore-from-disk, and serialized turn queue — one account's long turn no longer blocks another's.
- Prompt snapshot cache keyed by active home (per-uid souls → per-uid prompts).
- **Lazy per-user birth**: first signed-in request births that user's soul in the background (deduped per uid); chatting before it completes runs on the bare prompt and the soul arrives mid-conversation.
- Account deletion clears the home subtree + in-memory contexts.

## Mac edition: zero change

No scope is ever entered → `lisaHome()` degrades to the global home; the global chat context is byte-for-byte the old code path. Background schedulers (idle/reflect) stay global-only; per-uid autonomy is gated on per-uid budgets (B4+).

## Tests

New tenancy suite: scope resolution, await-survival, **two tenants writing disjoint souls, cross-tenant reads impossible**, isBorn scoping. Full suite **1038 pass / 0 fail**.

## Notes / follow-ups

- Firestore for accounts/balances (multi-instance unlock) is deferred to a follow-up PR — file-backed state stays correct under `min=max=1`, which deploy.sh still pins.
- GCS `/data` needs no change: per-uid is just the `/data/users/<uid>` subtree.

### DIFF
--- a/src/advisor/engine.ts
+++ b/src/advisor/engine.ts
@@ -11,7 +11,7 @@
 
 import fsp from "node:fs/promises";
 import path from "node:path";
-import { LISA_HOME } from "../paths.js";
+import { lisaHome } from "../paths.js";
 import { atomicWrite } from "../fs-utils.js";
 import { withFileLock } from "../soul/lock.js";
 import { runAllDetectors } from "./detectors.js";
@@ -30,7 +30,9 @@ const MIN_SCORE = 1.5; // relevance bar: below this → journal, not user
 const DIGEST_THROTTLE_MS = 3 * 60 * 60_000; // ≤1 non-urgent digest / 3h
 const RE_ARM_MS = 24 * 60 * 60_000; // a standing condition can re-surface daily
 
-export const ADVISOR_STATE_PATH = path.join(LISA_HOME, "advisor-state.json");
+export function advisorStatePath(): string {
+  return path.join(lisaHome(), "advisor-state.json");
+}
 
 /** Relevance score = urgency × actionability × dismissal-decay. */
 export function scoreSuggestion(s: Suggestion, state: AdvisorState): number {
@@ -133,7 +135,7 @@ export function applyDismissal(
 export async function dismissSuggestion(
   id: string,
   category: SuggestionCategory,
-  statePath: string = ADVISOR_STATE_PATH,
+  statePath: string = advisorStatePath(),
 ): Promise<void> {
   await withFileLock(statePath + ".lock", async () => {
     const state = await loadAdvisorState(statePath);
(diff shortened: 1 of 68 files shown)


oratis:
### 🤖 Automated review — ✅ MERGE (nits) — isolation empirically verified

The filesystem home-isolation boundary (this PR's charter) is **correct and verified by running the tests**, not just reading:
- `tenancy.test.ts` **6/6 pass** (disjoint soul subtrees, cross-tenant reads impossible, scope survives awaits).
- `enterWith` cross-request leak reproduced on Node 26: interleaved/keep-alive requests with `accountUid=null` correctly resolve to GLOBAL, never a prior request's uid.
- uid = `apple-<sanitized>` / `em-<hex>` — always one fs-safe segment, HMAC-protected in the session token → no `../`/absolute traversal.
- Lazy birth deduped via `birthsInFlight`.

Nits (non-blocking at reviewer scale):
- `soul/lock.ts:135` + `soul/git.ts:30` are the **only two module-level constants still frozen at import** (→ GLOBAL soul). Not a leak (soul *files* write per-uid), but every tenant serializes on one global lock and litters `.write.lock`/`.git-write.lock` in the global tree. Functionize (contradicts the PR's "all home-capturing constants functionized" note).
- `ctxForRequest` (`server.ts:340-350`) has a same-uid cold-start race → duplicate session; guard with an in-flight promise.
- `promptCache`/`uidChats` are unbounded (no eviction) — add LRU before real multi-user.

⚠️ **Out-of-scope but important** (pre-existing, NOT introduced here): the process-wide `moodBus`/`eventClients` broadcast (`server.ts:380-393`) fans `mood` + `chat_start/end` + `idle_message` **text** to **all** connected accounts regardless of uid — a genuine cross-user content leak. The SSE layer must be tenant-isolated before multi-user launch.

PR #263, opened 2026-07-21
feat(billing): B3 — price table + per-uid usage meter + usage API

oratis:
**Stacked on #262.** Milestone B3: metering. Interactive chat was already computing token usage every turn and **discarding it** at the `runAgent` result — this PR catches it, prices it, and ledgers it per user.

## What

- **`prices.ts`** — longest-prefix model table, integer micro-USD per M tokens. Face = provider list × **1.4 margin**. GLM = *standard* tier (free-window eligible in B4); Claude/GPT = *premium* (paid balance only). Unknown models fall back to premium pricing — a table gap can never hand out free inference. `PRICES_VERSION` stamps every ledger line.
- **`meter.ts`** — append-only `billing/usage.jsonl` in the ACTIVE home (per-uid under B2's scope): the audit source the B4 `balance.json` fast-path can always be rebuilt from. Same lock+trim discipline as `autonomy/runs.jsonl`; `recordUsage` never throws.
- **`server.ts`** — `/chat` meters each turn (cloud edition only; Mac/BYO stays unmetered), `GET /api/billing/usage` returns 12h-window + today summaries for the signed-in account.
- Fixed two module-scope paths in `autonomy/runs.ts` that still froze the home at import (missed in B2's sweep).

## Tests

Price tiers / margin math / round-up-never-free / ledger round-trip / per-home scoping / corrupt-line tolerance. Full suite **1045 pass / 0 fail**.

### DIFF
--- a/src/autonomy/runs.ts
+++ b/src/autonomy/runs.ts
@@ -21,8 +21,12 @@ import { withFileLock } from "../soul/lock.js";
 export function autonomyDir(): string {
   return path.join(lisaHome(), "autonomy");
 }
-const RUNS_FILE = path.join(autonomyDir(), "runs.jsonl");
-const RUNS_LOCK = path.join(autonomyDir(), "runs.lock");
+function runsFile(): string {
+  return path.join(autonomyDir(), "runs.jsonl");
+}
+function runsLock(): string {
+  return path.join(autonomyDir(), "runs.lock");
+}
 const MAX_RUNS = 2000;
 
 /** Which self-driven mechanism produced the run. */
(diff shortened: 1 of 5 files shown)


oratis:

- Price math verified exact (1M glm output = `face(2.2)` = 3,080,000 µUSD = $3.08). `costMicroUSD` = `Math.max(0, Math.ceil(...))` (never free / never negative). Unknown model → **premium** `FALLBACK` (a table gap can't hand out free/standard-tier inference). Usage API is home-scoped (no cross-tenant IDOR).

Nits (non-blocking):
- `prices.ts:79-87` — `costMicroUSD` has no finite guard: a NaN/Infinity token count → `NaN`, which poisons `summarizeUsage`/`balance.json`/the breaker. **Clamp non-finite → 0 (being added in the hardening commit).**
- `server.ts:1018` — `GET /api/billing/usage` doesn't gate on `accountUid` the way `/api/billing/quota` does; a shared-demo-token caller gets the GLOBAL home's aggregate usage. **Adding the same `{available:false}` guard in the hardening commit.**
- `prices.ts:66-72` — `priceForModel` is first-match, not longest-prefix; a future mis-ordered insert silently mis-prices. Sort by prefix length or add an order-guard test.

PR #264, opened 2026-07-21
feat(billing): B4 — 12h session window + paid tiers + 402 paywall + budget breaker

oratis:
**Stacked on #263.** Milestone B4 — the free-allowance / tier engine, per the locked decisions.

## Engine (`src/billing/quota.ts`)

- First standard-model turn opens a **12h window** with a tier-dependent allowance:

| Tier | Condition (30d cumulative IAP) | Window |
|---|---|---|
| Free | — (Apple / verified email) | $5 |
| Free-unverified | — (email, pre-verification) | $1 |
| Tier 1 | ≥ $4.99 | $10 |
| Tier 2 | ≥ $19.99 | $20 |

- Tier perk decays 30 days after purchases stop; **paid balance never expires** (3.1.1).
- Debit order: free window → paid balance. **Premium models (Claude/GPT) never touch the free window** — the one hard economics constraint from §7.
- Refund clawback may push paid negative: premium locks, standard free window survives (decision #4).

## Server

- `/chat` pre-checks **before** the SSE handshake → clean **HTTP 402** `{error, resetAt, tier}`; granted budget feeds `runAgent`'s `budgetTokens` breaker (every token priced at the output rate) so one turn can't overshoot; actual metered cost debited post-turn.
- `GET /api/billing/quota` — window/tier/balance for all clients (same numbers everywhere).
- Legacy shared-token demo stays ungated (operator-funded).

## iOS

- ChatView maps 402 → paywall copy; AccountCard gains the **session allowance bar** (spent/left + reset countdown), tier + credits rows.

## Tests

Window open/reset, tier math + 30d decay, debit order, premium isolation, clawback semantics. Full suite **1051 pass / 0 fail**; iOS simulator build green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #265, opened 2026-07-21
feat(billing): B5 — StoreKit 2 consumable credits + server JWS verify + ASN refunds

oratis:
**Stacked on #264.** Milestone B5 — the money path. Native StoreKit 2 + zero-dependency server verification (no RevenueCat).

## The invariant

**Credit exactly once, then finish.** The client sends the transaction JWS to the server *before* `Transaction.finish()`; the server verifies Apple's signature chain, dedupes by `transactionId` in a **global index** (one Apple transaction can never credit two accounts), credits the signed-in uid's balance, and only then does the client finish. A crash anywhere in between → StoreKit re-delivers → the dedup makes the replay a no-op.

## Server

- `verifyAppleJWS`: ES256 by the x5c **leaf**, chain walked with validity windows, root pinned to **Apple Root CA G3** (fetched once from apple.com over TLS, cached; `LISA_APPLE_ROOT_PATH` for air-gapped).
- `POST /api/billing/iap` (session-authed): validate bundle+product → credit → return fresh quota. Duplicate → `200 {ok:false}` so the client still finishes.
- `POST /api/billing/asn` (pre-gate — the signature chain IS the auth): **REFUND/REVOKE** → claw the credit back from the owning account (balance may go negative → premium locks per B4). Verified notifications always 200.

## iOS

`CreditsStore` (+ `Transaction.updates` listener at launch) and `PaywallSheet` (3 packs / restore / tier explainer); AccountCard gets **Add credits…**.

## Operator actions (before 1.1 review)

1. ASC: sign the **Paid Apps** agreement (banking/tax), create the 3 **consumable** products with the exact ids above, submit them WITH the 1.1 binary.
2. Apply to the **Small Business Program** (15%).
3. ASC → App Information → set the **ASN V2 production URL** to `https://<cloud>/api/billing/asn`.
4. Create + fund the review demo account (`reviewer@meetlisa.ai`) and put user/pass in App Review Information.

## Tests

JWS shape rejection / payload validation / credit + cross-account dedup / refund clawback + post-refund replay. Full suite **1056 pass / 0 fail**; iOS build green.

### DIFF
--- a/packaging/ios-companion/Sources/AccountViews.swift
+++ b/packaging/ios-companion/Sources/AccountViews.swift
@@ -156,6 +156,7 @@ struct AccountCard: View {
     @State private var showDeleteConfirm = false
     @State private var deleting = false
     @State private var quota: LisaClient.QuotaStatus?
+    @State private var showPaywall = false
 
     private static let tierLabels: [String: String] = [
         "free": "Free", "free-unverified": "Free (verify email for more)",
@@ -186,6 +187,11 @@ struct AccountCard: View {
                     }
                     LabeledContent("Tier", value: Self.tierLabels[q.tier ?? "free"] ?? (q.tier ?? "Free"))
                     LabeledContent("Credits", value: Self.dollars(max(0, q.paidMicroUSD ?? 0)))
+                    Button {
+                        showPaywall = true
+                    } label: {
+                        Label("Add credits…", systemImage: "plus.circle")
+                    }
                 } else {
                     LabeledContent("Plan", value: (acct.plan ?? "free").capitalized)
                 }
@@ -201,6 +207,11 @@ struct AccountCard: View {
             }
         }
         .task { quota = try? await app.client.billingQuota() }
+        .sheet(isPresented: $showPaywall, onDismiss: {
+            Task { quota = try? await app.client.billingQuota() }
+        }) {
+            PaywallSheet().environmentObject(app)
+        }
         .confirmationDialog("Delete your LISA account?", isPresented: $showDeleteConfirm, titleVisibility: .visible) {
             Button("Delete account and data", role: .destructive) {
                 deleting = true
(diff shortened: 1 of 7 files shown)


oratis:
### 🤖 Automated review — ⛔️ BLOCKER (fixable) — merged with the fix applied on `main`

**The JWS verification itself is sound** — I read the code: `verifyAppleJWS` does real ES256 `crypto.verify` over `header.payload` with the x5c leaf key, walks the cert chain (validity window + `cert.verify(issuer.publicKey)`) up to a pinned Apple Root G3. Signature forgery is not possible. Idempotency/refunds are **solid** (global `iap-transactions.json` under `withFileLock` dedupes cross-account; `clawbackPurchase` is idempotent; negative balance handled).

⛔️ **Blocker — StoreKit environment not enforced.** `validateTransaction` (`src/billing/iap.ts:158-169`) parses `environment` but never checks it; the server path (`server.ts:1077`) only logs it. **Sandbox** transactions are Apple-signed with the *same* G3 chain and cost the buyer **$0** → anyone with a free sandbox tester Apple ID can "buy" `credits.20`, POST the sandbox JWS to prod `/api/billing/iap`, pass signature+bundle+product+dedup, and receive **$22 of real credit**, repeatable with a fresh `transactionId`.
**Fix:** require `tx.environment === "Production"` on the prod server (accept `Sandbox` only when a `LISA_IAP_ALLOW_SANDBOX`/non-prod flag is set). This is applied in the post-merge hardening commit on `main`.

Further hardening (backlog):
- Chain walk doesn't assert `basicConstraints CA:TRUE`/`keyUsage=keyCertSign` on intermediates nor the leaf's Apple OID — prefer Apple's `app-store-server-library` path validation.
- The "pinned" root is fetched over TLS from apple.com and cached, not fingerprint-pinned — hardcode the G3 SHA-256 and assert it.
- `iap.ts:181-192` — global-index write and per-uid credit are in separate locks; a crash between them burns the `transactionId` with no credit (user silently loses the pack).

oratis:
✅ **Blocker fixed post-merge** in #273 (`main` @ 6923194): the cloud `/api/billing/iap` endpoint now rejects any non-`Production` StoreKit transaction (`sandbox_rejected`), with `LISA_IAP_ALLOW_SANDBOX=1` as the staging escape hatch. The sandbox-JWS free-credit path is closed. The deeper cert-path-validation and root-pinning hardening notes above remain on the backlog.

PR #266, opened 2026-07-21
feat(gateway): B6 — inference gateway + key-free signed-in Macs + lisa login/billing

oratis:
**Stacked on #265.** Milestone B6 — decision #3: **a signed-in local Mac needs no API key.**

## Gateway (`src/web/gateway.ts`)

Two protocol faces, both `POST`-only and account-session-only (the shared demo token is refused):

- `/gw/anthropic/v1/messages` → api.anthropic.com (x-api-key swap)
- `/gw/openai/v1/chat/completions` → the model's OpenAI-compat preset (GLM → open.bigmodel.cn)

Per request: **B4 quota precheck** (premium models need paid balance; exhaustion = structured 402) → key-swap → **streaming passthrough** with a tee-parser that reads token usage out of the stream itself (anthropic `message_start`/`message_delta`; openai via injected `stream_options.include_usage`) → metering + debit into the uid's ledger.

**Privacy boundary** (documented): prompts transit the gateway over TLS and are never persisted — the ledger stores token counts + model name only. Users who don't accept that keep the BYO-key path, unchanged.

## Managed mode (providers/registry.ts)

`lisa login` writes `LISA_MANAGED_SESSION`/`_BASE` to config.env; any model **without its own BYO key** resolves to the gateway (Anthropic face via `authToken`, OpenAI face via bearer). **BYO keys always win.** `hasCredentialsForModel` accepts a managed session so birth/run gates pass key-free. (Gemini has no gateway face yet.)

## CLI

- `lisa login [base]` — email+password (echo-off TTY; never argv), stores the session, prints the quota.
- `lisa logout` — removes it (BYO keys unaffected).
- `lisa billing` — tier / session window / credits / last-12h usage.

## Follow-ups noted

- Mac menu-bar app "Sign in" UI that fronts `lisa login` (the backend plumbing this PR adds is what it needs).
- Gemini gateway face if demand appears.

## Tests

Upstream planning + key swap, stream usage folding (both faces), non-stream usage, managed resolution + key gate. **1064 pass / 0 fail.**

### DIFF
--- a/src/cli-args.ts
+++ b/src/cli-args.ts
@@ -40,7 +40,10 @@ export interface ParsedArgs {
     | "sense"
     | "agents"
     | "pair"
-    | "mail";
+    | "mail"
+    | "login"
+    | "logout"
+    | "billing";
   subargs: string[];
   serveWeb: boolean;
   serveImessage: boolean;
@@ -190,7 +193,10 @@ export function parseArgs(argv: string[]): ParsedArgs {
       first === "sense" ||
       first === "agents" ||
       first === "pair" ||
-      first === "mail"
+      first === "mail" ||
+      first === "login" ||
+      first === "logout" ||
+      first === "billing"
     ) {
       out.subcommand = first;
       out.subargs = positional.slice(1);
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -201,6 +201,23 @@ async function main(): Promise<void> {
     return;
   }
 
+  // Managed inference accounts (PLAN_ACCOUNTS_BILLING B6).
+  if (args.subcommand === "login") {
+    const { cmdLogin } = await import("./cli/account.js");
+    await cmdLogin(args.subargs);
+    return;
+  }
+  if (args.subcommand === "logout") {
+    const { cmdLogout } = await import("./cli/account.js");
+    await cmdLogout();
+    return;
+  }
+  if (args.subcommand === "billing") {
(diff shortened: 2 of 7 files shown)


oratis:
### 🤖 Automated review — ✅ MERGE (nits) — auth + metering enforced

Verified there is **no unauthenticated or unmetered path**:
- `/gw/*` requires `cloud && accountUid` → 403 otherwise; `accountUid` is set **only** by a verified account session, so the shared `LISA_WEB_TOKEN`/device tokens (which leave `accountUid=null`) are correctly refused.
- `handleGateway` runs `precheckTurn` **before** the upstream call and `debitTurn` after, inside the per-uid `homeScope`.
- Keys are injected server-side only (never returned to client); client headers dropped except `anthropic-version`; upstream host hardcoded per face → no SSRF.

Nits (non-blocking):
- `gateway.ts:144` — `for await (const chunk of req)` reads the body **unbounded** → a signed-in client can OOM the process and forward huge prompts to the paid upstream. Add a byte cap → 413.
- `gateway.ts:202-204` — metering trusts upstream self-reported usage; if the usage chunk is missing (openai-compat preset ignoring `include_usage`, or a mid-stream error), `debitTurn(0)` → free inference. Add a floor/estimate from input bytes.
- `gateway.ts:162+204` — precheck→call→debit is TOCTOU; N concurrent premium requests each pass precheck and drive the balance negative (bounded only by client concurrency).

PR #267, opened 2026-07-21
feat(billing): B7 — rate limits, daily cap, kill switch, reviewer-account seed

oratis:
**Stacked on #266.** Final milestone B7 — abuse backstops + App Review automation. This completes the B0–B7 plan.

## Guards (`src/billing/limits.ts`)

- **Per-uid RPM** (sliding minute, default 20, `LISA_RPM_LIMIT`) — calibrated so farming a full window needs sustained max-rate traffic.
- **Global daily cap** (default $200 face, `LISA_DAILY_CAP_USD`) — persisted under the global home (restart-proof); exceeded → the whole cloud 402s `service_paused` until the UTC day rolls.
- **Kill switch** `LISA_BILLING_KILL=1` — instant pause of all metered inference.
- One `preflightLimits()` verdict consumed by **both** metered paths (`/chat` and `/gw/*`).
- Anomaly alert: any account crossing **$10 face/day** logs loudly (push-channel wiring is a follow-up).

## Reviewer-account seed — the 2.1 terminal fix

`LISA_REVIEWER_SEED="reviewer@meetlisa.ai:<password>"` on the cloud instance idempotently creates a **verified** email account topped up to **$20 (Tier 2, premium included)**. ASC's *Sign-in required* fields finally hold real credentials; the seed never rotates an existing password.

## Deploy

`deploy.sh` passes `LISA_REVIEWER_SEED` / `LISA_RPM_LIMIT` / `LISA_DAILY_CAP_USD` / `LISA_BILLING_KILL` through. Plan doc gains the B0–B7 → PR implementation log + operator checklist.

## Tests

RPM slide + per-uid isolation, daily-cap accumulate/rollover/persistence, kill-switch/429 verdicts. Full suite **1069 pass / 0 fail**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #268, opened 2026-07-22
feat(accounts): B8a — email verification (Resend); verified levels $1 → $5

oratis:
**Stacked on #267.** First of the code-leftover finishers from PLAN §10.

- **`mailer.ts`** — Resend REST (no deps). No `RESEND_API_KEY` → the link is logged instead of sent (dev/offline safe).
- **`accounts.ts`** — `beginEmailVerification` (token hash only on disk, 24h TTL, rotation) / `confirmEmailVerification` (constant-time, replay-dead). Seed helper dead-code cleanup.
- **`server.ts`** — register sends the mail; `GET /verify?token=` pre-gate landing page; `POST /api/auth/verify/resend`.
- **iOS** — unverified banner + resend button in AccountCard.
- Quota: no change needed — `tierFor` already keys off `verified`, so confirming levels the window $1 → $5 automatically.

**Operator env (when ready):** `RESEND_API_KEY` + verify `meetlisa.ai` as a sending domain in Resend; optionally `LISA_MAIL_FROM`. Until then registration still works — links appear in the Cloud Run log.

Tests: 6 new (round-trip/replay/expiry/rotation/kind-gating/mailer). **1075 pass / 0 fail**; iOS build green.

### DIFF
--- a/packaging/ios-companion/Sources/AccountViews.swift
+++ b/packaging/ios-companion/Sources/AccountViews.swift
@@ -157,6 +157,7 @@ struct AccountCard: View {
     @State private var deleting = false
     @State private var quota: LisaClient.QuotaStatus?
     @State private var showPaywall = false
+    @State private var resendMessage: String?
 
     private static let tierLabels: [String: String] = [
         "free": "Free", "free-unverified": "Free (verify email for more)",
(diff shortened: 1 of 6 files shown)


oratis:

- Verification token = `crypto.randomBytes(24)` (192-bit), stored **only** as SHA-256, constant-time compared, 24h TTL, single-use (cleared + `verified` guard blocks replay). Test asserts the raw token is never persisted.
- No `$5` self-bump: the raw token is emailed to `acct.email` only; neither `/register` nor `/verify/resend` returns it. `RESEND_API_KEY` from env (no key committed).

Nits (non-blocking):
- `server.ts:~222` — `verifyLinkFor` builds the link from the client-controlled `Host` header (host-header injection). Low impact, but pin a trusted `LISA_PUBLIC_BASE` for defense-in-depth.
- `GET /verify` (`server.ts:1014`) mutates state (marks verified) → email link-scanners/prefetchers can auto-consume the token. Acceptable (single-use, intended outcome), but a POST-confirm is more robust.
- A matched-but-expired token returns null without clearing the stale hash (harmless; re-begin is the only cleanup).

PR #269, opened 2026-07-22
feat(auth): B8b — Sign in with Apple on the web login page

oratis:
**Stacked on #268.** Second code-leftover finisher.

- **`cloudAuth.ts`**: `LISA_CLOUD_APPLE_WEB_SID` + `audienceForClient` — web tokens carry the **Services ID** as `aud`, native ones the bundle id; an unconfigured surface is rejected, never mis-verified.
- **`server.ts`**: `/api/auth/apple` accepts `{client:"web"}`; success now also pins the session cookie. New pre-gate `GET /api/auth/config` (feature flags + ids only) tells the login page what to draw.
- **`login.ts`**: Apple button appears only when configured; Apple JS popup → id_token → POST → reload into the authed island.

## Operator (to enable, optional)

1. Apple portal → Identifiers → **Services ID** (e.g. `ai.meetlisa.web`) → enable *Sign in with Apple*, group under primary App ID `ai.meetlisa.main`, verify domain `cloud.meetlisa.ai`, return URL `https://cloud.meetlisa.ai`.
2. Deploy with `LISA_CLOUD_APPLE_WEB_SID=ai.meetlisa.web`.

Until then the page shows email-only — no behavior change.

Tests: config parsing + audience selection. Suite green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #270, opened 2026-07-22
feat(billing): B8c — Stripe top-up (desktop/web) + /account page

oratis:
**Stacked on #269.** Third code-leftover finisher — the desktop/web money path.

- **`stripe.ts`** (zero deps): hosted Checkout via form-encoded REST, ad-hoc `price_data` (same face economics as the IAP packs), `uid`/`pack` in metadata; **Stripe-Signature v1** verification (timing-safe, 5-min tolerance); refund → session lookup by payment_intent.
- **`iap.ts`**: `creditExternalTransaction` — the global-dedup + scoped-credit core is now channel-agnostic; one transaction can never credit twice **across channels** either.
- **`server.ts`**: `POST /api/billing/stripe/checkout` (account session) · `POST /api/billing/stripe/webhook` (pre-gate, raw-body signature auth; replay-tolerant credit, refund clawback) · **`GET /account`** — signed-in self-service page (quota bar, credits, top-up, sign-out). Top-up buttons render only when Stripe is configured; **iOS never links here** (3.1.1).
- `deploy.sh`: passes `STRIPE_SECRET_KEY`/`STRIPE_WEBHOOK_SECRET` + the B8a mail envs that were missing.

## Operator (to enable, optional)

1. Stripe account → copy the secret key.
2. Dashboard → Webhooks → add `https://cloud.meetlisa.ai/api/billing/stripe/webhook`, events `checkout.session.completed` + `charge.refunded` → copy the signing secret.
3. Deploy with both envs. Until then `/account` simply hides the top-up section.

Tests: pack economics / checkout body / signature valid+tampered+stale / classification. Suite green.

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -66,6 +66,12 @@ ENVS="^##^LISA_EDITION=cloud##LISA_WEB_TOKEN=${LISA_WEB_TOKEN}"
 #   LISA_RPM_LIMIT / LISA_DAILY_CAP_USD  abuse guards (defaults 20 rpm / $200 per day)
 #   LISA_BILLING_KILL=1                  pause ALL metered inference immediately
 [ -n "${LISA_REVIEWER_SEED:-}" ]  && ENVS="${ENVS}##LISA_REVIEWER_SEED=${LISA_REVIEWER_SEED}"
+#   RESEND_API_KEY / LISA_MAIL_FROM  email verification (B8a; domain verified in Resend)
+#   STRIPE_SECRET_KEY / STRIPE_WEBHOOK_SECRET  desktop/web top-up (B8c)
+[ -n "${RESEND_API_KEY:-}" ]        && ENVS="${ENVS}##RESEND_API_KEY=${RESEND_API_KEY}"
+[ -n "${LISA_MAIL_FROM:-}" ]        && ENVS="${ENVS}##LISA_MAIL_FROM=${LISA_MAIL_FROM}"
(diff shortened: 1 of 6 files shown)


oratis:
### 🤖 Automated review — ✅ SAFE TO MERGE — Stripe path is correct

The two high-risk checks, verified in source:
- **Webhook signature IS verified, not trusted**: `server.ts:942-948` reads the raw body then gates everything on `verifyStripeSignature(raw, sig, secret)` (bad sig → 401, missing secret → 404). `stripe.ts` computes `HMAC-SHA256(\`${t}.${payload}\`)`, compares with `timingSafeEqual` after a length check, enforces a 300s timestamp tolerance, iterates multiple `v1` sigs (rotation-safe). A forged POST cannot mint credits.
- **Credit amount is server-side**: face value = `STRIPE_PACKS[pack].faceMicroUSD`; the charge is set server-side in `createCheckoutSession`; client supplies only a pack key. `uid` = session `accountUid`, never the client body → users can only top up their own account.
- **Idempotent**: dedup via `stripe-<sessionId>` in the global tx-index; replayed webhooks → `duplicate_transaction` → 200. `/account` is static HTML, data via `accountUid`-scoped APIs.

Nits (non-blocking):
- `server.ts:942` — `raw += chunk.toString("utf8")` per-chunk; a multibyte char split across chunks corrupts the body and fails an otherwise-valid signature. Accumulate Buffers and `Buffer.concat` before verifying (fails closed → reliability, not security).

PR #271, opened 2026-07-22
feat(mac): B8d — menu-bar Sign in to LISA Cloud + anomaly push alerts

oratis:
**Stacked on #270.** Fourth code-leftover finisher.

## Mac menu-bar app

- **`AccountWindow`** (Lisa ▸ *Sign in to LISA Cloud…*): email+password → session written to `~/.lisa/config.env` (`LISA_MANAGED_SESSION`/`_BASE`) → this Mac runs **key-free** through the B6 gateway. BYO keys always win; Sign out clears the session. In-window **Restart backend to apply** (pkill the detached `serve --web`, fresh start, outcome via the existing status notification).
- `BackendController` gains public `configEnvValue`/`upsertConfigEnv` (replace-or-append, 0600 kept) + `restart()`.

## Server

- `meter.setAnomalySink` + `PushBridge.onBillingAnomaly`: the **$10/day per-account anomaly** now reaches the operator's phone via the existing push channel (pref "error", bridge-throttled) instead of only the Cloud Run log.

Verified: mac-client `build.sh` green; typecheck + full suite green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #272, opened 2026-07-22
feat(cloud): B9 — Firestore state backend + turn lease; multi-instance behind LISA_FIRESTORE=1

oratis:
**Stacked on #271.** The last code leftover — and the end of the B-plan. **Everything defaults OFF**: without `LISA_FIRESTORE=1` the file backends under `min=max=1` behave exactly as before.

## What moves to Firestore (all CAS, zero deps)

| State | Doc | Why it blocked multi-instance |
|---|---|---|
| Account directory | `lisa-global/accounts` (one doc, email-uniqueness checked *inside* the CAS) | file read-modify-write races |
| Per-uid balance | `lisa-balances/{uid}` | debit/credit races |
| Purchase dedup | `lisa-txindex/{txId}` create-only (`exists:false`) — dedup becomes a **datastore property**, valid across instances *and* channels | per-instance file lock |
| Daily cap | `lisa-global/day-YYYY-MM-DD` + 30s read cache | per-instance counter |
| Turn serialization | `lisa-leases/turn-{uid}` — chat + gateway acquire before running; wait ≤15s then **429 turn_in_progress**; TTL 180s backstops crashes | per-instance ChatCtx chain only |

## Client (`src/cloud/firestore.ts`)

ADC token from the metadata server, JSON⇄Firestore codec, doc ops with `currentDocument` preconditions, `casUpdate` (retry-on-contention), leases. `accounts.ts` API became **async** behind the seam (call sites + tests updated).

## Deploy

`MAX_INSTANCES` knob; the script **refuses** `>1` without `LISA_FIRESTORE=1`. Runbook gains the enable checklist (enable API, Native-mode DB, `datastore.user` role, one-time accounts import — ask for the import script when switching).

## Tests

Codec round-trip · preconditions · **CAS retry under a sabotaged race** · lease acquire/busy/renew/release/expiry-takeover · accounts/verification suites on the async API. **1089 pass / 0 fail.**

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -72,6 +72,10 @@ ENVS="^##^LISA_EDITION=cloud##LISA_WEB_TOKEN=${LISA_WEB_TOKEN}"
 [ -n "${LISA_MAIL_FROM:-}" ]        && ENVS="${ENVS}##LISA_MAIL_FROM=${LISA_MAIL_FROM}"
 [ -n "${STRIPE_SECRET_KEY:-}" ]     && ENVS="${ENVS}##STRIPE_SECRET_KEY=${STRIPE_SECRET_KEY}"
 [ -n "${STRIPE_WEBHOOK_SECRET:-}" ] && ENVS="${ENVS}##STRIPE_WEBHOOK_SECRET=${STRIPE_WEBHOOK_SECRET}"
+#   LISA_FIRESTORE=1  move accounts/balances/tx-index/day-cap/turn-lease to Firestore
+#   (enable the API + Native-mode DB in the project first; unlocks MAX_INSTANCES>1)
+[ -n "${LISA_FIRESTORE:-}" ]         && ENVS="${ENVS}##LISA_FIRESTORE=${LISA_FIRESTORE}"
+[ -n "${LISA_FIRESTORE_PROJECT:-}" ] && ENVS="${ENVS}##LISA_FIRESTORE_PROJECT=${LISA_FIRESTORE_PROJECT}"
 [ -n "${LISA_RPM_LIMIT:-}" ]      && ENVS="${ENVS}##LISA_RPM_LIMIT=${LISA_RPM_LIMIT}"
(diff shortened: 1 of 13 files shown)


oratis:
### 🤖 Automated review — ✅ MERGE (nits) — CAS + lease atomicity verified

Everything is behind `LISA_FIRESTORE=1` (default off); file paths unchanged. The critical concurrency checks, verified in source:
- **CAS is genuinely atomic** (not read-then-write): `casUpdate` reads `{fields, updateTime}`, applies a pure fn, commits with a `currentDocument` precondition (`updateTime`, or `exists:false` for creates); a racing writer → 409/412 → retry with backoff. Firestore enforces the precondition server-side.
- **Lease mutual exclusion holds**: `acquireLease` refuses a live lease held by another owner; two instances racing an expired/absent lease both hit the same precondition → exactly one commits, the loser re-reads the winner's lease and returns busy. Crash-safe via `expiresAt` TTL; `releaseLease` only clears if `owner` still matches. Tests cover acquire/busy/renew/release/expiry-takeover.
- `scopedUid()` rejects path separators — load-bearing since admin creds bypass Firestore rules by design; it's correctly derived from the verified session.

Nits (non-blocking):
- `turn-lease.ts:19` — `TURN_TTL_MS = 180_000` but chat SSE runs under `--timeout 3600` with **no lease renewal/heartbeat**. A metered turn exceeding 180s lets its lease expire mid-flight; a second concurrent same-uid turn on another instance can then acquire → the double-debit / racing-soul-write the lease exists to prevent (only under `LISA_FIRESTORE=1` + `MAX_INSTANCES>1`). Add periodic renewal or raise TTL toward the turn timeout. **Raising TTL in the hardening commit.**
- `limits.ts` `globalSpendExceeded` is a 30s-cached read + fire-and-forget add; the $200 global cap can overshoot one refresh window (cold cache returns 0). Acceptable for an abuse guard with the ledger authoritative.
- `firestore.ts:606` `void parts;` is dead code; `releaseLease`'s `.replace(/^lisa-leases\//,"lisa-leases/")` is a no-op — cleanup.

PR #278, opened 2026-07-23
docs(kb): knowledge base v2.0 plan — ingest · daily brief · link graph

oratis:
Design of record for the three knowledge-base capabilities requested on top of v1.0 (#235–#242).

## What it plans

| | Capability | New modules |
|---|---|---|
| **K1** | Paste any link (web / 公众号 / B站 / YouTube) → Markdown source with provenance | `src/kb/ingest/` |
| **K2** | A watchlist → daily fetch → classify → summarized brief (push + stored as knowledge) | `src/kb/feeds/` |
| **K3** | `[[slug]]` parsed into a real link graph; `index.md` becomes a ranked map-of-content; memory stores links, not content | `src/kb/links.ts` |

## Review of the shipped v1.0 code — five gaps, one of them a live bug

**Confirmed bug (G5):** both TF-IDF tokenizers (`src/kb/search.ts:45`, `src/memory/vector.ts:212`) split on whitespace. CJK has none, so an entire Chinese run collapses into one token:

```
doc  "这篇公众号文章讲的是知识库的设计" → ["这篇公众号文章讲的是知识库的设计"]
query "知识库"                        → ["知识库"]        → 0 hits
```

Chinese content is only findable by an exact whole-string query. Fixed in K-B with CJK bigrams, before any Chinese content is ingested.

The rest:
- **G1** no ingest path — `web_fetch` returns lossy plain text (`htmlToText` flattens every heading/list/link), no metadata, no dedupe, no store.
- **G2** no KB scheduler — mail is the only daily job in the repo.
- **G3** `normalizeSlug` strips non-ASCII, so every Chinese title falls back to `entry-<timestamp>`.
- **G4** `index.md` is a flat regenerated TOC; `[[slug]]` is documented in `SCHEMA.md` but no code reads it — no backlinks, no orphans, no graph.

## Notable decisions (full debate in §6)

- **Self-built extractor**, not a remote reader — `r.jina.ai` would send every article the user reads to a third party, which breaks the 100%-local promise.
- **Video transcripts are best-effort, never a success condition** — YouTube's PO Token and Bilibili's `SESSDATA` login requirement are dependencies that will keep breaking; tiering is built-in API → optional `yt-dlp` → metadata-only.
- **Autonomous ingest is a prompt-injection path** (external page → `index.md` → every system prompt). Three-way containment: watchlist-only URLs for autonomous runs, `origin: web` sources contribute title-only to the index, untrusted-content fencing in `kb_read`.
[...]

PR #279, opened 2026-07-23
fix(kb): CJK-safe search + slugs, and provenance frontmatter (K-B)

oratis:
Foundation for the KB v2 work — [plan](../blob/feat/kb-v2-plan/docs/PLAN_KNOWLEDGE_BASE_v2.0.md) §K-B. Three fixes that would otherwise corrupt or lose data once link ingest starts pouring Chinese content in, so they land before the firehose.

Stacked on #278.

## 1. CJK search was broken (live bug)

Both TF-IDF tokenizers — `src/kb/search.ts:45` and `src/memory/vector.ts:212` — carried the same three lines: lowercase, strip non-word chars, **split on whitespace**. Chinese has no whitespace, so a whole clause survived as one token:

```
doc   "这篇公众号文章讲的是知识库的设计"  →  ["这篇公众号文章讲的是知识库的设计"]
query "知识库"                          →  ["知识库"]
intersection = ∅                        →  0 hits
```

Chinese content was findable only by an exact, whole-clause query — which nobody types. This affected the knowledge base *and* `memory_search` over transcripts.

Fixed by a shared `src/tokenize.ts`:
- CJK runs emit character **bigrams** plus the run itself (so an exact phrase still ranks highest). Bigram indexing is the standard dictionary-free approach for CJK; IDF handles bigrams that straddle word boundaries.
- Runs are segmented at latin↔CJK boundaries, so `gpt模型` indexes as `gpt` + CJK bigrams instead of one unmatchable blob.
- Kana and CJK Ext-A are now kept — the old character class was `[一-鿿]` only, so Japanese text tokenized to *nothing*.
- Latin tokenization is byte-identical to before; `memory/vector.ts` keeps its extra transcript stopwords by passing a superset.

## 2. Chinese titles produced throwaway slugs

`normalizeSlug` does `replace(/[^a-z0-9]+/g, "-")`, so a CJK title normalized to `""` and `addSource` fell back to `entry-<Date.now()>` — opaque, unstable, impossible to dedupe against.

New `src/kb/slug.ts`: latin titles slug exactly as before; anything with fewer than 3 usable latin chars gets `<date>-<sha256/8>`. Slugs stay **ASCII on purpose** — macOS stores filenames NFD and git/Linux NFC, so a CJK filename written on one and read on the other is a different byte string; readability is the `title:` frontmatter's job, and every surface shows the title anyway.

`canonicalUrl()` strips `utm_*` and known tracking params (bilibili `spm_id_from`/`vd_source`, wechat `srcid`/`scene`, youtube `si`/`feature`) so the same article shared from two places hashes identically. Deliberately keeps `t`, `from`, `timestamp` — those carry real meaning on some sites, and this canonical URL is also what gets stored as the entry's `url:`, so stripping a meaningful param would hand the user back a broken link.

## 3. Provenance frontmatter

`KbEntry.extra` round-trips arbitrary frontmatter keys. Ingest provenance (`url`/`site`/`author`/`published`/`hash`/`via`) lives there without the store knowing about any of it — and a key a future version adds survives a round-trip through an older one instead of being silently dropped. Values are whitespace-collapsed on write so a newline in an extracted author name can't forge frontmatter lines on the next read (tested).

## Tests

+32 tests, all green (`1158 pass / 0 fail` full suite, typecheck clean). Notably `src/kb/search.test.ts` is an end-to-end regression for the bug above: a `知识库` query against a Chinese source returns it as the top hit.

### DIFF
--- a/docs/PLAN_KNOWLEDGE_BASE_v2.0.md
+++ b/docs/PLAN_KNOWLEDGE_BASE_v2.0.md
@@ -1,0 +1,543 @@
+# PLAN — 知识库 v2.0：摄取 · 日报 · 链接图
+
+> 目标：把 v1.0 的"能存能搜的三层 KB"升级成"**会自己长大的知识系统**"——
+> ① 每天自动抓取并总结成信息日报；② 粘贴任意链接（公众号 / 网站 / B站 / YouTube）
+> 一键转成带出处的 Markdown 知识；③ 所有知识用 index 织成链接图，memory 通过 link 调取。
+>
+> 参考：[docs/PLAN_KNOWLEDGE_BASE_v1.0.md](PLAN_KNOWLEDGE_BASE_v1.0.md)（已 shipped #235–#242）、
+> [src/kb/](../src/kb/)、[src/mail/](../src/mail/)（日报调度的现成范式）、[src/prompt.ts](../src/prompt.ts)。
+>
+> **Status:** 提案（待评审） · **日期:** 2026-07-23 · **Author:** Claude (for Oratis)
+> 本文含**现状 review（含 1 个已确认 bug）+ 完整设计 + 正反方辩论 + 分阶段 PR 计划**。
+
+---
+
+## 0. TL;DR — 三个能力，一句话各是什么
+
+| # | 能力 | 一句话 | 核心新模块 |
+|---|---|---|---|
+| **K1** | 链接 → Markdown | 粘贴任意链接 → 带出处 frontmatter 的 Layer-1 source | `src/kb/ingest/` |
+| **K2** | 抓取 + 信息日报 | 一份 watchlist → 每日拉取 → 分类总结 → 日报（推送 + 落库） | `src/kb/feeds/` |
+| **K3** | 索引链接图 | `[[slug]]` 真正被解析成图；index 变 MOC；memory 用 link 调取 | `src/kb/links.ts` |
+
+三者不是三个独立功能，而是**一条流水线**：K2 发现 → K1 摄取 → K3 织网 →
+idle 蒸馏（v1.0 已有）→ 长成 wiki。K3 是黏合剂，也是最该先做的（见 §7 排序理由）。
+
+---
+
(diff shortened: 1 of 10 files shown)


oratis:
Good fix, well-tested — the shared bigram tokenizer is the right dictionary-free approach and the `search.test.ts` regression (`知识库` → top hit) nails the reported bug. Latin parity holds byte-for-byte. Provenance round-trip is injection-safe on the value side (verified by the newline test).

Non-blocking items:

1. **CJK wiki upsert across days.** `writeWiki`'s auto-mint path uses `kbSlug({title})` = `<today>-<hash(title)>`, so the same Chinese concept title written on a later day mints a *different* slug and creates a duplicate page. Latin titles and the explicit-`slug` path are fine. Consider dropping the date component for the wiki layer (concepts are upserted, not chronological captures).

2. **Dedupe story isn't wired yet.** `canonicalUrl` is exported/tested but unused; `addSource` hashes the raw `extra.url` with a capture-time date and stays append-only (`uniqueSlug` → `-2`), so re-ingesting the same link does not dedupe here. Fine if K1 owns dedupe, but the "same link → same slug" docstring oversells the current behavior — worth a canonicalize-then-hash in `addSource`.

3. **Harden `extra` keys.** `serializeEntry` sanitizes values against newline-forgery but not keys. Not reachable today (keys are code-set), but since `extra` is a pass-through bag, validate keys against `^[a-zA-Z_][a-zA-Z0-9_]*$` on write to keep it safe if a future caller ever derives a key from untrusted data.

Nit: `kbSlug(...) || \`page-${Date.now()}\`` in `writeWiki` is now unreachable — `kbSlug` never returns "".

oratis:
Pushed the frontmatter **title whitespace-collapse + extra-key validation** fix (6d1d28d) — this closes the untrusted-title → frontmatter/prompt-injection issue at its root in `store.ts`, so the ingest (#283) and adapters (#284) titles inherit it. +2 regression tests; typecheck + 42 KB tests green. Merging.

PR #280, opened 2026-07-23
feat(kb): parse [[links]] into a real graph; index.md becomes a ranked MOC (K-C)

oratis:
[Plan](../blob/feat/kb-v2-plan/docs/PLAN_KNOWLEDGE_BASE_v2.0.md) §K-C. Stacked on #279.

## The gap

v1.0 told Lisa to cross-link wiki pages with `[[slug]]` (`src/kb/schema.ts:20`) and recorded each page's `sources:` in frontmatter — but **nothing in the repo ever read either one**. The KB was a pile of files with decorative links: no backlinks, no way to see what a page connects to, no way to tell a hub from an orphan.

## What lands

**`src/kb/links.ts`** builds the graph from three edge sources — `[[slug]]`, `[[slug|alias]]`, `[[kb:slug]]` in bodies, plus a wiki page's `sources:` frontmatter — and derives what search can't give you: backlinks, hubs, orphans, and links pointing at pages that don't exist yet. A bare `[[slug]]` resolves wiki-first, since a link written in prose means the concept page, not a raw capture that happens to share the slug.

**`index.md` is now a ranked map of content**, not a flat listing. It's injected into every system prompt and hard-capped there (`src/prompt.ts:165`), so with a flat list the truncation lands at an arbitrary point once the KB grows — and whatever fell below the line stops existing as far as Lisa is concerned. Ordering by `(1 + backlinks) × recency` means the cut always eats the least-connected tail. Orphans and broken links are listed as the wiki's to-do list (what the idle tending pass acts on), and every truncation states how much it dropped rather than trailing off silently.

**Sources are listed by title only.** Layer 1 is raw captured material — once link ingest lands, that means whole web pages — and `index.md` is always-on prompt text, so an excerpt there is a direct *any page on the internet → Lisa's system prompt* path. This is a stronger form of the containment the plan assigned to K-I (it covers every source, not just `origin: web`), and it costs nothing: a title is all a map needs and the body is one `kb_read` away. Wiki gists stay — Lisa wrote those herself. There's a test that feeds an injection string through a source body and asserts it never reaches the index.

**Also:**
- `kb/index.json` — the machine-readable graph, for the web UI and tools.
- `kb_links(slug)` — links to / linked from / shares tags with. Read-only, remote-safe.
- `kb_read` now takes a bare slug or a `[[wikilink]]` with `layer` optional (the model routinely passes wikilinks verbatim), and appends the pages that link to what you just read.

## Tests

+27 tests. Full suite `1184 pass / 0 fail`, typecheck clean. Notable: hub ordering puts a stale orphan last, an injection string in a source body never reaches `index.md`, and a 40-page/300-source KB still renders under 8 KB.


oratis:
The graph model is clean (pure `buildGraph`, no traversal so cycles are inert) and the title-only source containment is the right call — nicely tested. One blocker plus a couple of wiring issues.

**Blocker — ReDoS in `WIKILINK` (`src/kb/links.ts`).** Catastrophic backtracking, measured:

    "[[" + " ".repeat(3000)  -> 6.1 s
    "[[" + " ".repeat(5000)  -> 25.6 s
    "[[" + " ".repeat(10000) -> did not finish (>2 min)

(tabs trigger it too; `"a"*3000` is 0 ms.) The greedy leading/trailing `\s*` overlap the whitespace-including lazy class `[^\]|#\n]+?`, so a `[[` with a long whitespace run and no closing `]]` blows up. This is not hypothetical: `parseWikilinks` runs over **every entry body, including `sources`**, inside `buildGraph`, which fires on every KB mutation *and* every `kb_read`/`kb_links`. `addSource` bodies are already arbitrary captures today and become whole web pages after K1 — one crafted or whitespace-heavy page bricks the KB subsystem. Note: bounding the class alone is **not** enough (I tested it — the trailing `\s*` still hangs on 100 K). Drop the surrounding `\s*` (you already `.trim()` the capture) and bound the lengths:

    const WIKILINK = /\[\[(?:kb:)?([^\]|#\n]{1,200}?)(?:\|[^\]\n]{0,200})?\]\]/g;

Verified: 0 ms on 100 K spaces, and identical output on every existing case (dedup, `|alias`, `kb:`, non-spanning, `[[ spaced ]]` → `spaced`). Please add a pathological-input test.

**Perf — `kb_read` is now O(N).** It calls `listFullEntries()` + `buildGraph()` on every read (all files, full bodies, target read twice) purely to show backlinks — and re-runs the parser above over everything. Cache the graph by KB-dir fingerprint the way `search.ts` caches its index. `regenerateIndexLocked` also builds the graph twice (once in `renderIndex`, once for `index.json`) — build once, pass both.

**Wiring — the to-do list falls off the bottom.** Orphans/broken render after up to 40 wiki + 24 tags + 25 sources. `renderIndex` targets ~8 KB, but the idle pass caps `index.md` at 1200 chars (`idle/runner.ts:149`) and the live prompt at 2600 (`prompt.ts:163`), so orphans/broken — the exact thing the idle "tend the wiki" pass is told to act on — get sliced off in any real KB. Move those sections above sources, or align `renderIndex`'s budget with the caps.

Minor: `[[slug#anchor]]` currently matches nothing (class excludes `#`, but nothing consumes the anchor) — decide whether to resolve to the base slug or report it. And source *titles* (attacker-controlled for web ingests) still reach `index.md` verbatim even though bodies don't — worth length-capping them.

oratis:
Pushed the **ReDoS fix** (8e10da0): dropped the surrounding `\s*` and length-bounded the target/alias — 0ms on 100k chars, byte-identical output on all existing cases, +regression test. typecheck + 18 links tests green. Merging. (The O(N) `kb_read`/double-graph-build and orphans/broken index ordering remain as noted — good follow-ups, not blockers.)

PR #296, opened 2026-07-24
docs(plan): web signup, login & the signed-in app v1.0 (S-series)

oratis:
## What

Design doc for the next era: **meetlisa.ai signup/login + the signed-in cloud app**, answering four requirements — Google/Apple/email+OTP login, automatic post-signup birth, the signed-in app with the full web feature set, and GCP (Cloud Run + GCS) production deployment.

## Highlights

- **Dual-path invariant (owner directive)**: self-hosted local (no account, flagship, forever) **and** optional cloud — the user keeps the power. Every design below must keep both paths visible and migratable.
- Gap analysis against the already-landed B-series (accounts/billing, PRs #259–#267): Apple + email/password + lazy per-uid birth + Cloud Run/GCS **already exist**; the real gaps are Google sign-in, email OTP (+ the missing password reset), birth transactionality/ceremony, per-uid autonomy scheduling, the website entry point, and productionization.
- Seven For/Against debates with verdicts (brand vs cloud account, Google reversal, custom zero-dep auth vs hosted IdP, OTP vs link, "full features" honest definition, ceremony vs silent birth, single-instance vs Firestore).
- Milestones **S1–S7**; S6 (Firestore + prod project + Secret Manager) is the hard gate before public signups.

Implementation PRs follow, one per milestone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #297, opened 2026-07-24
feat(auth): Sign in with Google on the web login page (S1)

oratis:
Part of [PLAN_WEB_SIGNUP_v1.0](https://github.com/oratis/LISA/pull/296) — milestone **S1**.

## What

Adds the third sign-in provider to LISA Cloud's login page, mirroring the Apple channel's zero-dependency posture:

- **`src/web/googleAuth.ts`** — pure RS256/JWKS verifier for Google ID tokens: dual issuer spellings, `aud` = OAuth web client id, clock tolerance, 1h-cached JWKS, typed `GoogleAuthError`. Default-OFF: lights up only when `LISA_CLOUD_GOOGLE_SIGNIN` **and** `LISA_CLOUD_GOOGLE_CLIENT_ID` are both set.
- **`POST /api/auth/google`** — pre-gate access-minting endpoint (same posture as `/api/auth/apple`), sharing the per-IP auth rate bucket; upserts the account, mints the per-uid HMAC session, pins the cookie.
- **Merge policy** (plan §3, D2 verdict): a Google sign-in whose email Google asserts verified links onto an existing **verified** same-email account — one uid, one Lisa across providers. Unverified squatter accounts never inherit a Google identity. Covered by tests.
- **Login page**: GIS button drawn only when `/api/auth/config` advertises `googleWeb`; script-load failure (e.g. mainland China) silently hides the button instead of erroring.
- `LOGIN_HTML` added to the inline-script syntax regression guard (it wasn't covered).
- `deploy/deploy.sh` plumbs the two new env vars.

## Replay posture

GIS button-flow tokens are minted per page-load by Google's script with no client-mintable raw nonce, so the ~1h token lifetime is the replay window — the same stance the Apple channel takes for pre-nonce clients. Documented in the module header.

## Tests

`npm test`: 1296 pass / 0 fail. New: `googleAuth.test.ts` (verifier + config, incl. alg-confusion and tampered-payload cases), 4 merge-policy cases in `accounts.test.ts`.

Create the OAuth web client in GCP console (authorized JS origin `https://cloud.meetlisa.ai`), then deploy with `LISA_CLOUD_GOOGLE_SIGNIN=1 LISA_CLOUD_GOOGLE_CLIENT_ID=…`.

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -61,6 +61,12 @@ ENVS="^##^LISA_EDITION=cloud##LISA_WEB_TOKEN=${LISA_WEB_TOKEN}"
 # Sign in with Apple on the WEB login page (B8b): the Services ID registered in
 # the Apple portal for cloud.meetlisa.ai (needs domain verification there).
 [ -n "${LISA_CLOUD_APPLE_WEB_SID:-}" ] && ENVS="${ENVS}##LISA_CLOUD_APPLE_WEB_SID=${LISA_CLOUD_APPLE_WEB_SID}"
+# Sign in with Google on the WEB login page (S1, PLAN_WEB_SIGNUP):
+#   LISA_CLOUD_GOOGLE_SIGNIN=1     enable POST /api/auth/google
+#   LISA_CLOUD_GOOGLE_CLIENT_ID=…  OAuth web client id (…apps.googleusercontent.com;
+#                                  authorized JS origin = https://cloud.meetlisa.ai)
+[ -n "${LISA_CLOUD_GOOGLE_SIGNIN:-}" ]    && ENVS="${ENVS}##LISA_CLOUD_GOOGLE_SIGNIN=${LISA_CLOUD_GOOGLE_SIGNIN}"
+[ -n "${LISA_CLOUD_GOOGLE_CLIENT_ID:-}" ] && ENVS="${ENVS}##LISA_CLOUD_GOOGLE_CLIENT_ID=${LISA_CLOUD_GOOGLE_CLIENT_ID}"
 # Accounts & billing era (PLAN_ACCOUNTS_BILLING B1–B7), all optional:
 #   LISA_REVIEWER_SEED="email:password"  idempotent App-Review demo account (verified, $20/Tier-2)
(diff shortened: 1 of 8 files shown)


oratis:
## Review — merge-ready after one fix ⚠️ (adversarial pass)

The two things that must be right in an auth foundation are both **correct**: ID-token verification (RSA signature verified *before* any claim is trusted, `alg` pinned to RS256 so `alg:none`/HS-confusion are dead, `kid` from Google's JWKS over TLS, `iss`/`aud`/`exp`/`iat` all checked) and the **account-takeover defense** — a Google login only merges onto an existing account when **both** Google asserts `email_verified` **and** the local account is already `verified:true` (reached only via the mailed-token flow). Every takeover path reduces to "already controls the email." Session minting is byte-identical to the password path. No client secret in `deploy.sh` (none needed). Nicely done.

**MUST-FIX — login-CSRF (Medium).** `readJsonBody` (`server.ts`) never checks `Content-Type` and tolerantly `JSON.parse`s any body. `POST /api/auth/google` has no `state`/nonce and no Origin check. An attacker submits their **own** valid Google credential cross-site via `<form enctype="text/plain">` (a CORS "simple request" → no preflight); the server sets `lisa_token=<attacker session>` in the victim's jar, and on the victim's next same-site visit they silently operate **inside the attacker's account** (session fixation → their chats/uploads/spend land in the attacker's Lisa). `SameSite=Strict` doesn't help (it governs sending, not storing).

**Fix I'll push:** enforce `Content-Type: application/json` centrally in `readJsonBody` (forms can't forge it → forces a preflight that fails without CORS). One change covers `/api/auth/google` + the shared siblings (`apple`, `login`, and S2's new OTP routes). Optional GIS `nonce` (mirror the Apple `expectedNonce` from #261) as follow-up.

Non-blocking: `verified: emailVerified !== false` defaults a *missing* claim to true — prefer `=== true`; JWKS unknown-kid could refresh-once before rejecting.

oratis:
Superseded by the A-series auth that landed on main via #305 (`Land auth OTP + Google + mail A1–A6`) — main already has Google sign-in + OTP, so this now conflicts on 7 files with no new content. The two review findings that were LIVE on main (login-CSRF via readJsonBody, and — for #301 — the birth crash-window) are fixed in #306. Closing as superseded per the plan; reopen if any piece here isn't actually covered by the A-series.

PR #300, opened 2026-07-24
fix(mail): detect real imapflow auth rejections; stop blaming the password for host errors

oratis:
Found by running the connect flow end-to-end against the real
`imap.gmail.com:993` after #234 shipped.

## Bug 1 — the app-password hint never fired (the common case)

imapflow reports an auth rejection as a bare **`message: "Command failed"`**;
the real signal is on the error *object*:

```
message             : "Command failed"
authenticationFailed: true
serverResponseCode  : "AUTHENTICATIONFAILED"
responseText        : "Invalid credentials (Failure)"
```

`friendlyMailError()` matched only on `message`, so a wrong Gmail password —
the single most common failure, and the exact thing the guided modal exists to
prevent — fell through to the generic branch and showed:

> Could not connect: Command failed

Now it keys on `authenticationFailed` / `serverResponseCode` / `responseText`
too, and the generic fallback prefers the server's own words over imapflow's
opaque `"Command failed"`.

## Bug 2 — a bad hostname was reported as a credentials problem

Caught by the new tests. The loose auth regex contained a bare `invalid`, and a
DNS error embeds the hostname:

```
getaddrinfo ENOTFOUND imap.nope.invalid
                                ^^^^^^^ matched the auth branch
```

So a typo'd IMAP host told the user their password was wrong, sending them off
to regenerate an app-password that was never the problem. Transport failures
(DNS / timeout / refused) are now checked **before** the text heuristic, and
`invalid` is dropped (`credential` already covers "Invalid credentials").

Resulting precedence: structured auth flags → transport → text heuristic → generic.

## Testability

Extracted the mapper from `server.ts` into `src/mail/connect-error.ts` so it is
unit-testable without importing the whole web server, and added
`connect-error.test.ts` with the **real** imapflow error shape captured from a
live probe pinned as a fixture — that shape is subtle enough that it needs a
regression test.

## Verification

Live, against the deployed backend:

```
# wrong Gmail password → real imap.gmail.com rejection
Gmail rejected the sign-in. Use a 16-character app password (not your Google
login password), and make sure 2-Step Verification is on.   [HTTP 401]

# bad IMAP host → transport hint, not a password hint
Could not reach the mail server (network or timeout). …      [HTTP 401]
```

Neither failed attempt was persisted (verify-before-store still holds).
`tsc --noEmit` clean · full suite **1288 pass / 0 fail**.

### DIFF
--- a/src/mail/connect-error.test.ts
+++ b/src/mail/connect-error.test.ts
@@ -1,0 +1,67 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { friendlyMailError } from "./connect-error.js";
+
+/**
+ * The exact error imapflow throws when Gmail rejects an app-password, captured
+ * from a live imap.gmail.com:993 probe. Note `message` is a useless "Command
+ * failed" — an earlier version keyed only on that and so fell through to the
+ * generic branch, which defeated the whole point of the guided connect UI.
+ */
+function gmailAuthRejection(): Error {
+  return Object.assign(new Error("Command failed"), {
+    response: "3 NO [AUTHENTICATIONFAILED] Invalid credentials (Failure)",
+    responseStatus: "NO",
+    executedCommand: "3 AUTHENTICATE PLAIN",
+    responseText: "Invalid credentials (Failure)",
+    serverResponseCode: "AUTHENTICATIONFAILED",
+    authenticationFailed: true,
+  });
+}
+
+test("real imapflow Gmail auth rejection → the app-password hint, not the generic branch", () => {
+  const msg = friendlyMailError(gmailAuthRejection(), "me@gmail.com", "imap.gmail.com");
+  assert.match(msg, /16-character app password/);
+  assert.match(msg, /2-Step Verification/);
+  assert.doesNotMatch(msg, /Command failed/);
+});
+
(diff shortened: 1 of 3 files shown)


oratis:
## Review — merge-ready ✅

Delivers its goal cleanly: no host/network/TLS error is ever blamed on the password. Classification is **safe-by-ordering** — structured auth flags (`authenticationFailed`, `serverResponseCode==="AUTHENTICATIONFAILED"`) first, then transport syscall codes (DNS, then network/timeout) *before* the loose text heuristic (so a hostname like `imap.nope.invalid` in an ENOTFOUND message isn't misread as auth), then an honest fallback surfacing the server's own `responseText` (not the useless "Command failed"). Traced every host/network/cert class through it — none misclassify as a password failure, and auth failures don't leak into the host buckets. Robust to null/undefined/string input; **no password/secret leak** (only message/responseText/code feed the output; `executedCommand`/full response are withheld). Old inline copy fully removed.

Non-blocking polish (I'll fold in): a dedicated TLS/cert branch (cert errors currently hit the generic fallback), a few more network codes (`EHOSTUNREACH`/`ENETUNREACH`), fold `err.cause` into the signal set, and two coverage cases (serverResponseCode-only, code-based transport).

PR #302, opened 2026-07-24
feat(cloud): per-uid autonomy sweep + Mac-bridge entry (S4)

oratis:
Part of [PLAN_WEB_SIGNUP_v1.0](https://github.com/oratis/LISA/pull/296) — milestone **S4**. **Stacked on #301** (S3) → #299 → #297.

## What

On a Mac, Lisa's idle/reflect schedulers run beside their one user. On the cloud they only ever ticked the **global** scope — a signed-in tenant's soul never reflected, never grew. Since soul/desires/reflection are LISA's differentiation, this was the biggest honest gap in "全部 web 功能" (plan §4.4, D5 verdict: the cloud's completeness = the soul system complete, machine-bound powers via the Mac bridge).

- **`autonomy-sweep.ts`** — walks accounts active in the last 7 days and, inside each uid's home scope, runs one reflection over that user's latest session (the REVE-lite tick). Cadence is **tier-gated**: free 24h · tier1 6h · tier2 1h — autonomy is simultaneously cost-capped and a reason to pay. `maxRuns` (default 20) bounds a single sweep's LLM spend; the per-uid stamp lands only after a successful reflection so failures retry next tick; per-user try/catch keeps one bad home from blocking the rest.
- **`POST /internal/autonomy/sweep`** — cloud-only, default-OFF without `LISA_SWEEP_TOKEN`, constant-time bearer check, returns the sweep report. Cloud Scheduler job snippet documented in `deploy.sh` (`*/30 * * * *` suggested; the per-tier stamps make ticks idempotent).
- **Account page**: honest "Your Mac" section — what cloud Lisa carries (soul, memory, knowledge, chat) vs what stays on the user's machine (PTY/dispatch/sense), with the pairing bridge signposted per the dual-path invariant.

## Tests

`npm test`: 1324 pass / 0 fail. New: 5 sweep cases (activity window, unborn skip, stamp gating via a real dreamFn-born soul, tier cadence ordering, per-user failure isolation).

Set `LISA_SWEEP_TOKEN` at deploy and create the Cloud Scheduler job (snippet in deploy.sh).

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -79,6 +79,12 @@ ENVS="^##^LISA_EDITION=cloud##LISA_WEB_TOKEN=${LISA_WEB_TOKEN}"
 [ -n "${LISA_TURNSTILE_SECRET:-}" ]   && ENVS="${ENVS}##LISA_TURNSTILE_SECRET=${LISA_TURNSTILE_SECRET}"
 # Extra disposable-email domains to block at signup (comma-separated; S3).
 [ -n "${LISA_EMAIL_BLOCKLIST:-}" ]    && ENVS="${ENVS}##LISA_EMAIL_BLOCKLIST=${LISA_EMAIL_BLOCKLIST}"
+# Per-uid autonomy sweep (S4): bearer secret for POST /internal/autonomy/sweep.
+# Pair it with a Cloud Scheduler job, e.g.:
+#   gcloud scheduler jobs create http lisa-autonomy-sweep --schedule "*/30 * * * *" \
+#     --uri "$URL/internal/autonomy/sweep" --http-method POST \
+#     --headers "Authorization=Bearer $LISA_SWEEP_TOKEN" --location "$REGION"
+[ -n "${LISA_SWEEP_TOKEN:-}" ]        && ENVS="${ENVS}##LISA_SWEEP_TOKEN=${LISA_SWEEP_TOKEN}"
 # Accounts & billing era (PLAN_ACCOUNTS_BILLING B1–B7), all optional:
 #   LISA_REVIEWER_SEED="email:password"  idempotent App-Review demo account (verified, $20/Tier-2)
 #   LISA_RPM_LIMIT / LISA_DAILY_CAP_USD  abuse guards (defaults 20 rpm / $200 per day)
--- a/src/reflect.ts
+++ b/src/reflect.ts
@@ -5,6 +5,7 @@ import { DEFAULT_MODEL } from "./llm.js";
 import { appendMemory, type MemoryStore } from "./memory/store.js";
 import { reflectionsDir } from "./paths.js";
 import { providerForModel } from "./providers/registry.js";
+import type { ProviderUsage } from "./providers/types.js";
 import { createSkill, getSkill, patchSkill } from "./skills/manager.js";
 import { withSoulCaller } from "./soul/git.js";
 import { recordAutonomyRun } from "./autonomy/runs.js";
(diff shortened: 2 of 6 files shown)


oratis:
## Review — request changes 🔴 (adversarial pass)

Isolation is **sound** (`homeScope.run(homeForUid(uid), …)` scopes every read/write per uid; one corrupt soul can't affect another — test-verified) and the new `POST /internal/autonomy/sweep` is **properly gated** (cloud-only + `LISA_SWEEP_TOKEN`, default-OFF→404, timing-safe bearer, no uid param so you can't target another user). The "Mac-bridge entry" is just static marketing copy, no endpoint. `account-page.ts` has zero interpolation — no XSS. But:

**MUST-FIX 1 — the sweep spends *premium* inference outside every billing gate (policy call for you).** `sweepOne` → `reflectOnSession` → `runTurn` on the premium default `claude-sonnet-4-6` with **no `precheckTurn`, no `debitTurn`, no `recordUsage`**. So: free/unverified users (who by rule have *no* premium access) get premium Sonnet; the spend is never charged; and it's **invisible to the $200/day cap, not stopped by `LISA_BILLING_KILL`, and absent from the audit ledger** — the one inference path that escapes both operator backstops. Same class as #301-M2 and #276-M1. Minimum: route through kill-switch + `recordUsage` (I can do); entitlement/debit policy is yours.

**MUST-FIX 2 — `maxRuns` has no server ceiling.** `body.maxRuns` accepted as any int → one authed call with `maxRuns: 1e9` sweeps every born user in a single request. **Fix I'll push:** clamp (`Math.min(n, 100)`).

Non-blocking: same idle session re-reflected every interval (~168 redundant premium runs/week — I'll add a "session advanced since last stamp" guard), wrap the handler body in try/catch (a Firestore `loadAccounts` throw → hung request), respond 202 + detach so a >180s sweep doesn't overlap on Scheduler retry.

oratis:
## Adversarial review + rebase onto `main` (billing floor applied)

Rebased S4's commit onto current `main` (S3 #301 now merged) — **zero conflicts**; deploy.sh and server.ts auto-merged cleanly beside S3's Turnstile/birth-hub wiring. typecheck clean; full suite **1383 pass / 0 fail**. Base retargeted `feat/s3-birth-hardening` → `main`.

### Findings & resolutions

**1. Billing floor — autonomy (M2, high).** The sweep's `reflectOnSession` fires premium inference through `withSoulCaller` → the provider **directly**, bypassing the metered turn path (`gateway.ts`/`server.ts` → `preflightLimits` + `recordUsage`). So autonomy spent with **no kill switch, no $200/day cap, and no audit line**. Per the "免费 + 底线" decision, autonomy stays free to the user but gains a floor:
- **Pre-check** `killSwitchOn() || globalSpendExceeded(now)` before each reflection — trips ⇒ the sweep stops and reports `service_paused` (the next scheduled tick resumes).
- **`recordUsage("autonomy", …)` after** each reflection — **no `debitTurn`** (the user is never charged), but the face cost is audited to `usage.jsonl`, counted against the global daily cap, and fed to the anomaly alert. `reflect.ts` now returns `usage` (the reflector+retry token totals it already tracked) so the cost is exact.

**2. Billing floor — birth (M2, carried from S3 #301).** Birth's dream is the same unmetered `withSoulCaller` path. Added the same `killSwitchOn() || globalSpendExceeded()` pre-check at the lazy-birth call site (`ensureUserBirth`): paused ⇒ birth defers and the soul is born on a later login once billing resumes (chat meanwhile runs with the bare prompt — the existing not-yet-born behavior). *Follow-up:* threading `usage` out of `birth.ts` for exact birth-cost metering (birth is once-per-account and now gated by single-flight + Turnstile + disposable + IP-cap, so the pre-check is the load-bearing guard).

**3. `maxRuns` clamp (correctness).** The caller's `maxRuns` reached the loop uncapped — a misconfigured Cloud Scheduler value could spend the day's inference in one tick. Clamped to `MAX_SWEEP_RUNS = 100`, with `NaN`/negative collapsing to the default 20.

**4. Handler try/catch (robustness).** The `/internal/autonomy/sweep` route called `sweepUserAutonomy` with no try/catch — a sweep-wide throw (e.g. an unreadable accounts store; per-uid failures are already absorbed) would hang the request instead of answering the scheduler. Wrapped to return `500 sweep_failed` cleanly, matching sibling routes.

**5. Test added.** `billing floor: the kill switch halts the sweep before any inference` — asserts `LISA_BILLING_KILL=1` yields `service_paused` (not the `unborn` the account would otherwise get), proving the floor runs ahead of any spend.

PR #304, opened 2026-07-24
feat(deploy): production hardening — Secret Manager mode, B9 import script, prod runbook (S6)

oratis:
Part of [PLAN_WEB_SIGNUP_v1.0](https://github.com/oratis/LISA/pull/296) — milestone **S6**, the hard gate before public signups. **Stacked on #302** (S4) → #301 → #299 → #297.

### `SECRETS_MODE=sm` for deploy.sh
The deploy script passed every secret (LLM keys, Stripe keys, the web token…) as **plain `--set-env-vars`**, visible to anyone with console read access. In `sm` mode those values are now pushed to **Secret Manager** (secret created if absent, new version per deploy, runtime SA granted accessor) and reach the container via `--set-secrets`. Default `env` mode keeps today's demo behavior byte-identical. All plumbing routes through one `addenv` helper — verified with a bash unit test of the routing and `bash -n`.

### The missing B9 import script
RUNBOOK_ACCOUNTS_LAUNCH's Firestore section literally said "问我要导入脚本" — flipping `LISA_FIRESTORE=1` on an empty database would orphan every existing account. `scripts/import-accounts-firestore.ts` imports the three file-backed stores into the **exact** docs the runtime seam reads:
- `accounts.json` → `lisa-global/accounts {list}`
- `users/<uid>/billing/balance.json` → `lisa-balances/<uid>`
- `iap-transactions.json` → `lisa-txindex/<txId>` (create-only, so replays are naturally deduped)

Idempotent; `--dry-run` / `--force`. `firestore.ts` gains a `LISA_FIRESTORE_TOKEN` override so admin tooling can authenticate with `gcloud auth print-access-token` outside Cloud Run (no metadata server there).

### `docs/RUNBOOK_CLOUD_PROD.md`
The operator checklist from reviewer-demo to public signup: dedicated prod project → Firestore create + import → `SECRETS_MODE=sm LISA_FIRESTORE=1 MAX_INSTANCES=3` deploy → third-party console table (Google OAuth client, Apple Services ID, Turnstile widget, Resend domain, Stripe webhook) → `cloud.meetlisa.ai` domain mapping (DNS-only/grey-cloud, TLS by Google) → Cloud Scheduler sweep job → uptime check + budget alerts + the built-in anomaly log alert → **three kill switches** → a launch smoke checklist. Every credential-gated step is flagged for the owner.

## Tests

`npm test`: 1324 pass / 0 fail; `bash -n` + an addenv routing check on deploy.sh.

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -52,57 +52,78 @@ if [ -z "${LISA_MODEL:-}" ]; then
   fi   # else: leave unset → Anthropic default (claude-sonnet-4-6)
 fi
 
-# Build the env list with a custom '##' delimiter (gcloud's ^d^ syntax) so secret
-# values may safely contain commas.
-ENVS="^##^LISA_EDITION=cloud##LISA_WEB_TOKEN=${LISA_WEB_TOKEN}"
-[ -n "${LISA_MODEL:-}" ]        && ENVS="${ENVS}##LISA_MODEL=${LISA_MODEL}"
-[ -n "${ANTHROPIC_API_KEY:-}" ] && ENVS="${ENVS}##ANTHROPIC_API_KEY=${ANTHROPIC_API_KEY}"
-[ -n "${ZHIPU_API_KEY:-}" ]     && ENVS="${ENVS}##ZHIPU_API_KEY=${ZHIPU_API_KEY}"
-[ -n "${OPENAI_API_KEY:-}" ]    && ENVS="${ENVS}##OPENAI_API_KEY=${OPENAI_API_KEY}"
+# ── Secret routing (S6) ─────────────────────────────────────────────────────
+# SECRETS_MODE=env (default): everything ships as plain env vars — unchanged
+#   demo behavior, values visible in the Cloud Run console.
+# SECRETS_MODE=sm: sensitive values go to Secret Manager (a version is pushed
+#   on every deploy) and reach the container via --set-secrets; only
+#   non-sensitive config stays in env vars. The PRODUCTION setting.
+SECRETS_MODE="${SECRETS_MODE:-env}"
+SECRET_VARS=" LISA_WEB_TOKEN ANTHROPIC_API_KEY ZHIPU_API_KEY OPENAI_API_KEY RESEND_API_KEY STRIPE_SECRET_KEY STRIPE_WEBHOOK_SECRET LISA_TURNSTILE_SECRET LISA_SWEEP_TOKEN LISA_REVIEWER_SEED "
+PENDING_SECRETS=()
+SET_SECRETS=""
+# Build the env list with a custom '##' delimiter (gcloud's ^d^ syntax) so
+# values may safely contain commas. addenv routes each var to env or secret.
+ENVS="^##^LISA_EDITION=cloud"
+addenv() { # addenv NAME VALUE
+  local name="$1" val="$2"
+  [ -z "$val" ] && return 0
+  if [ "$SECRETS_MODE" = "sm" ] && [[ "$SECRET_VARS" == *" $name "* ]]; then
+    PENDING_SECRETS+=("${name}=${val}")
+    SET_SECRETS="${SET_SECRETS:+$SET_SECRETS,}${name}=${name}:latest"
+  else
+    ENVS="${ENVS}##${name}=${val}"
+  fi
+}
+addenv LISA_WEB_TOKEN        "${LISA_WEB_TOKEN}"
+addenv LISA_MODEL            "${LISA_MODEL:-}"
+addenv ANTHROPIC_API_KEY     "${ANTHROPIC_API_KEY:-}"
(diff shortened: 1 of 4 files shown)


oratis:
## Review — request changes 🔴 (adversarial pass)

Secret Manager routing is **right** (every sensitive var covered, pushed via stdin not argv, removed from `--set-env-vars` so `gcloud run describe` shows only `secretKeyRef`s, no `set -x`/`echo $SECRET`; session secret is stable across deploys — sessions survive the cutover). The tx-index migration is idempotent/replay-safe and account fields round-trip losslessly. But the account/balance migration + runbook have real hazards:

**MUST-FIX 1 — a malformed/missing `accounts.json` silently writes `{list:[]}` → every account orphaned.** `readJson` swallows *all* errors → `null` → coerced to `[]`; on a first import the else-branch writes an empty list, then the runbook flips `LISA_FIRESTORE=1` and everyone is locked out. The runbook's own `gcloud storage rsync` is exactly what can half-populate the local dir. **Fix I'll push:** distinguish parse-failure from a genuinely-empty file and abort; refuse a zero-length accounts write without an explicit `--allow-empty`.

**MUST-FIX 2 — `--force` clobbers prod accounts + balances (money) with the stale snapshot, no CAS/backup/confirm.** `setDoc` with no `updateTime` precondition = full-document overwrite. A `--force` re-run post-cutover erases every account created after the snapshot and rolls `paidMicroUSD` back to snapshot values. **Fix I'll push:** CAS `--force` on the `updateTime` read (refuse if prod changed), snapshot-first + typed confirmation.

**MUST-FIX 3 — runbook sweep-token wiring is broken.** `LISA_SWEEP_TOKEN="$(openssl rand -hex 24)"` is a command-scoped prefix, never exported → the Scheduler `Authorization: Bearer $LISA_SWEEP_TOKEN` expands empty → 401 every tick → the S4 autonomy sweep never runs; and it rotates each deploy. **Fix I'll push:** generate once, persist + reuse.

Non-blocking but real: `deploy.sh` prints `?token=$LISA_WEB_TOKEN` (the master bearer) to stdout **even in `sm` mode** → the master key lands in CI/terminal logs, undercutting the whole PR; I'll gate that echo to `env` mode. Also: no prod-target guardrail (defaults to the shared personal project + `SECRETS_MODE=env`), session HMAC key lives in the GCS bucket (bucket-read = session forgery).

oratis:
## Adversarial review + rebase onto `main`

Rebased S6's commit onto current `main` (S3+S4 merged). One deploy.sh conflict: S6 refactors the env wiring into an `addenv` helper (with Secret Manager routing) while `main` carries the A-series Google vars inline — resolved to **S6's `addenv` style + the A-series `LISA_GOOGLE_{WEB,IOS}_CLIENT_ID`**, dropping S1's dead `LISA_CLOUD_GOOGLE_*` (0 readers). Verified the merged script has every var (A-series Google, S3 Turnstile/blocklist, S4 sweep, billing) once, no duplicates. `bash -n` clean; typecheck clean; full suite **1383 pass / 0 fail**.

### Findings & resolutions

**1. Import script — empty source wipes Firestore (data loss, high).** `scripts/import-accounts-firestore.ts` reads `accounts.json` as `… ?? []`, so a wrong/missing home dir (or an unmounted GCS bucket) yields `[]` — and with `--force` it would `setDoc("lisa-global/accounts", { list: [] })`, **erasing every account**. Added an empty-guard: `accounts.length === 0` ⇒ skip with a warning (an empty source is a misinvocation, not an import). The existing-doc guard + `--force` semantics are unchanged for real data.

**2. Runbook — sweep token never reaches the scheduler.** §2 generated `LISA_SWEEP_TOKEN=…` **inline** on the deploy command, so it lived only for that one process; §5's `gcloud scheduler jobs create … Bearer $LISA_SWEEP_TOKEN` and the verify `curl` then ran in a fresh shell with the var **unset** → the scheduler would authenticate with an empty bearer and every tick would 401. Fixed: `export LISA_SWEEP_TOKEN="$(openssl rand -hex 24)"` as its own step so both the deploy and §5 share it. Same pass fixed the runbook's **dead S1 `LISA_CLOUD_GOOGLE_*`** → A-series `LISA_GOOGLE_WEB_CLIENT_ID` / `LISA_GOOGLE_IOS_CLIENT_ID` (deploy block + the third-party-console table).

**3. deploy.sh — master token leaked into prod logs.** The final line echoed `…/?token=$LISA_WEB_TOKEN` unconditionally, printing the master token into the deploy log even under `SECRETS_MODE=sm` (production, where accounts — not the shared token — are the real auth). Gated: `sm` mode prints the plain console URL; the token-in-URL convenience stays for the `env` demo mode only.

`firestore.ts`'s `LISA_FIRESTORE_TOKEN` operator override (so the import runs outside Cloud Run) is clean and carried as-is.

PR #308, opened 2026-07-25
Enforce Cloud capability boundaries

oratis:
## What changed

- introduce explicit `local-owner` and `cloud-chat` capability profiles
- use a strict allow-list for hosted tools; unknown plugin, MCP, and future tools are denied by default
- skip executable skills, plugins, hooks, MCP connections, voice tools, and the full-tool `task` closure when booting the Cloud edition
- apply the filtered toolset to chat, idle runs, managed agents, and `/api/tools`
- deny host-control and arbitrary-outbound routes at the Cloud HTTP boundary, including agent/PTY control, dispatch, config, coding plans, device pairing, IMAP, screen/sense capture, and URL ingestion
- add denial-path tests for tools and routes

## Why

The Web server previously received the same assembled toolset as the local CLI. In a multi-tenant Cloud deployment that made local-owner capabilities such as shell, arbitrary file access, process control, operator plugins, and MCP tools reachable from a tenant-facing runtime. Client-side edition hiding and approval prompts are not authorization boundaries.

## Impact

Local and remote-device behavior in the Mac edition is unchanged. Hosted chat retains tenant-scoped Soul, memory, mood, and bounded KB tools. Cloud-only endpoints that control the host or initiate arbitrary network connections now return a structured 403 `capability_denied` response.

## Validation

- `npm run typecheck`
- `npm test`: 1,390 passed, 1 PTY-only skip, 0 failed
- `npm run build`
- focused capability tests: 13 passed
- `git diff --check`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -20,6 +20,7 @@ import { parseArgs, type ParsedArgs } from "./cli-args.js";
 import { connectMcpServers } from "./mcp/client.js";
 import { loadMcpConfig } from "./mcp/config.js";
 import { lisaHome } from "./paths.js";
+import { isCloud } from "./edition.js";
 import { loadAllPlugins, PLUGINS_ROOT } from "./plugins/loader.js";
 import type { HookSpec } from "./plugins/types.js";
 import { buildSystemPromptSnapshot, getPromptFingerprint } from "./prompt.js";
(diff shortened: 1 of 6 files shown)


oratis:
Code review: reviewed the hosted capability boundary end to end. The allow-listed cloud tool profile and server-side route denial are the right fail-closed shape. I found one host-configuration gap: /api/plans/select was not covered by the exact /api/plans denial. Fixed in 8f79a93 with a regression assertion; focused capability tests and TypeScript checks pass. No remaining blocking issue.

PR #309, opened 2026-07-25
Pin public origin and trusted proxy identity

oratis:
## What changed

- require a canonical HTTPS `LISA_PUBLIC_ORIGIN` when the Cloud edition starts
- build verification email and Stripe checkout URLs exclusively from that configured origin
- encode verification tokens through the URL API
- ignore `X-Forwarded-For` by default and add explicit right-to-left resolution through `LISA_TRUST_PROXY_HOPS`
- configure the Cloud Run deploy path with the production origin and one trusted proxy hop
- document both settings in English and Chinese
- add focused tests for hostile origins, token encoding, spoofed forwarding chains, invalid hops, and fail-closed fallback

## Why

Verification and checkout links were derived from request `Host` and `X-Forwarded-Proto`, allowing a hostile or misconfigured proxy header to place a raw verification token on an attacker-controlled origin. Authentication rate limits also selected the first forwarded address, which a caller could prepend and rotate.

## Impact

Cloud deployments must provide a valid HTTPS public origin; the bundled deployment script defaults it to `https://cloud.meetlisa.ai`. Local mode uses a deterministic loopback origin unless explicitly configured. Proxy headers affect rate-limit identity only when the trusted hop count is configured.

## Validation

- isolated committed snapshot: `npm run typecheck`
- isolated committed snapshot: focused origin/proxy tests, 8 passed
- isolated committed snapshot: `npm run build`
- full working-tree suite: 1,391 passed, 1 PTY-only skip, 0 failed
- deploy script syntax: `bash -n deploy/deploy.sh`
- `git diff --check`

Unrelated concurrent coding-plan changes in the shared worktree were deliberately excluded from this commit.


oratis:
Code review: canonical public-origin construction and right-to-left trusted-proxy resolution remove request-header control from verification, checkout, and rate-limit identity. After #308 merged, I resolved the server initialization conflict by retaining both the cloud capability profile and canonical-origin boundary. TypeScript checks plus 13 focused origin/proxy/capability tests pass. No remaining blocking issue.

PR #311, opened 2026-07-25
Harden billing state and payment recovery

oratis:
## Summary

- make account, balance, purchase, and global-spend state fail closed on corruption or storage outages
- persist Apple transaction progress through `pending -> credited -> refund_pending -> refunded`
- make balance credits idempotent by external transaction ID and safely resume crash-window retries
- return retryable 503 responses when durable billing state is unavailable
- retain compatibility with legacy credited transaction records

## Why

The previous storage helpers frequently treated unreadable or malformed durable state as an empty store. In billing paths that can recreate accounts, reset balances, duplicate credits, or silently bypass global spend controls. The new behavior distinguishes missing state from unavailable/corrupt state and makes every externally retried payment transition recoverable.

## Validation

- `npm test` — 1392 passed, 1 skipped
- `npm run build` — passed
- added coverage for corrupt stores, duplicate/conflicting credits, crash recovery, refund replay, pending-owner conflicts, and fail-closed spend cache behavior

## Rollout

No eager data migration is required. Existing valid account/balance records remain readable; legacy transaction-index entries without an explicit status are interpreted as `credited`. Deploy with normal billing-state backups and monitor `billing_state_unavailable` responses during rollout.


oratis:
Code review: reviewed fail-closed account/balance/index parsing, pending→credited and refund_pending→refunded recovery, transaction ownership conflicts, balance-level idempotency, and HTTP retry semantics. Found that Firestore transport/ADC failures could still surface as non-retryable verification errors; fixed in 3feda5b and added an outage regression test. Integrated cleanly with the latest cloud capability/autonomy main. TypeScript, build, and 75 focused tests pass. No remaining blocking issue.

PR #312, opened 2026-07-25
Bound every buffered HTTP request body

oratis:
## Summary

- replace every remaining unbounded `for await ... body += chunk` request reader in the web server
- use a shared 1 MiB ceiling for control-plane JSON
- allow a separately bounded 20 MiB ceiling for base64 audio and chat attachments
- return structured 413 responses and close overflowed connections
- return a deterministic 400 when the request stream becomes unavailable

## Why

Twenty-four authenticated and local routes still buffered arbitrarily large or endless bodies before parsing. Authentication limits who can trigger the behavior, but does not protect process memory from compromised clients, leaked device tokens, or accidental oversized uploads.

## Compatibility

This deliberately preserves each route's existing JSON parsing and media-type behavior. It adds only byte ceilings and consistent transport errors; valid existing clients require no changes.

## Validation

- repository-wide scan finds no executable unbounded request-body loops
- `npm test` — 1384 passed, 1 skipped
- `npm run build` — passed
- body-reader unit tests cover UTF-8 chunk boundaries, exact-limit acceptance, overflow, stream failure, and cap tiers


oratis:
Code review: reviewed the shared byte-counted reader, control-vs-rich request caps, structured 413/400 responses, and every migrated route. Repository-wide scan found three additional unbounded HTTP listeners outside src/web/server.ts (Feishu, Slack, generic webhook); all now use the same 1 MiB control-plane reader in 8403461, with an oversized Feishu regression test. The executable-source scan is clean, TypeScript/build pass, and focused body/channel/integration tests pass. No remaining blocking issue.

PR #310, opened 2026-07-25
feat: make Lisa's desires evolve with conversation, time, and web research

oratis:
## What changed

- adds the full v2 design and rollout plan in `docs/PLAN_DESIRE_DYNAMICS_v2.0.md`
- extends desires with intensity, horizon, semantic update/review timestamps, and bounded source URLs
- computes time-decayed effective intensity (`spark` / `season` / `enduring`) and uses it for current-desire ranking while preserving live conversation focus
- extends conversation reflection so it can add/revise the new dynamics fields
- fixes `soul_patch(field="desire")` updates resetting a desire's birth time
- adds a dedicated `desire_revise` tool that preserves identity and records review provenance
- adds a daily `builtin:desire_review` heartbeat that selects one due desire and may perform bounded web research
- constrains desire review to a narrow toolset, with hard limits of one search and two fetches per run
- fences `web_search` and `web_fetch` output as untrusted external content
- records desire review as its own autonomy-run kind and refreshes long-lived prompts on a daily desire clock

## Why

The v1 mechanism made conversation reflection capable of revising/closing desires, but time itself did not change a desire and background browsing was only an incidental tool opportunity. Activity mtime could also dominate the displayed desire, while same-slug `soul_patch` updates destroyed birth-time semantics.

This PR unifies conversation changes, temporal cooling/reinforcement, and bounded background evidence gathering in one auditable desire state. Web content can inform a desire but cannot grant shell, file mutation, dispatch, GitHub, MCP, or redeploy capabilities.

## User impact

Lisa's surfaced wants can now change as a conversation evolves and as short-lived interests cool over time. When Proactive heartbeat is enabled, she can privately review one due desire, search the web within a strict budget, and confirm, cool, transform, or close it with source provenance and progress history.

Cloud sweep parity and UI visibility are deliberately left as the separately planned PR C.

## Validation

- `npm run typecheck`
- `npm test` — 1406 passed, 1 environment-dependent skip, 0 failed
- `npm run build`
- focused desire-review suite — 5 passed
- `git diff --check`

PR #313, opened 2026-07-25
feat: bring desire review to cloud autonomy

oratis:
## What changed

- extracts a dedicated `runDesireReviewOnce` path that runs only the bounded desire-review builtin
- uses a per-tenant review lock based on the active `homeScope`
- guarantees a no-change review still persists `lastReviewedAt` when the model forgets to call the tool
- adds a cloud reflection cursor keyed by session id and user-message count
- prevents cloud sweep from repeatedly reflecting an unchanged transcript
- lets an otherwise-idle cloud cadence review one due desire even when there is no chat session
- limits each account to one inference action per sweep: new conversation reflection first, desire review otherwise
- meters review tokens through the existing cloud autonomy ledger and preserves kill-switch/tier/max-run gates
- passes the server's tool registry into the tenant-scoped sweep

## Why

The first PR completes the local desire dynamics loop, but the hosted sweep previously only reflected the newest session and could revisit the same transcript on every paid-tier cadence. It also had no path for time-driven desire review or bounded web research when a tenant had no new conversation.

This stacked PR brings the same review semantics and capability boundary to cloud accounts without doubling inference spend per sweep.

## Safety and cost

- one account gets at most one LLM action per sweep
- desire review retains the hard one-search/two-fetch budget and narrow tool allowlist from the base PR
- all work runs inside the tenant's `homeScope`
- Proactive off, billing kill switch, global spend cap, tier cadence, and `maxRuns` still gate execution
- review usage is recorded with zero-assumed cache tokens rather than left unmetered

## Validation

- `npm run typecheck`
- `npm test` — 1409 passed, 1 environment-dependent skip, 0 failed
- `npm run build`
- focused cloud sweep + desire review suites — 14 passed
- `git diff --check`

## Stack

Depends on #310. Base this PR on `main` after #310 merges.


oratis:
Code review: reviewed the tenant-scoped desire-review path, no-change review stamping, reflection cursor, one-action-per-account rule, cloud usage metering, and the narrow review tool boundary. Rebased the stack onto main after #310 and verified TypeScript, build, and 29 focused tests. This is safe to merge, with #314 next in sequence to add cross-instance checkpoint/lease idempotency for the shared sweep path.

PR #314, opened 2026-07-25
Make cloud autonomy sweeps idempotent

oratis:
## Summary

- serialize per-tenant autonomy work in-process and, when enabled, across instances with a Firestore lease
- persist atomic `pending` / `completed` checkpoints containing the reflected session and message count
- skip unchanged sessions even when the tier cadence becomes due again
- delay recovery of crash-left `pending` work until its lease window expires
- fail closed on corrupt checkpoint state instead of silently repeating reflection side effects

## Why

The cadence stamp limited frequency but did not identify what had already been reflected. Concurrent scheduler deliveries could reflect the same tenant at the same time, and an unchanged session was reprocessed every cadence, repeating model spend and potentially appending duplicate journal/memory operations.

## Recovery semantics

Normal failures restore the previous checkpoint and retry on the next scheduler tick. A hard process crash leaves a `pending` checkpoint; peers wait for the 30-minute lease backstop before retrying. This provides single-active-worker and unchanged-input idempotency while retaining at-least-once recovery for a crash inside the reflection operation itself.

## Validation

- `npm test` — 1387 passed, 1 skipped
- `npm run build` — passed
- new tests cover unchanged-input deduplication, appended-session eligibility, overlapping sweeps, and corrupt-checkpoint fail-closed behavior


oratis:
Code review: reviewed local and Firestore sweep serialization, pending/completed recovery semantics, corrupt-checkpoint fail-closed behavior, and unchanged-input deduplication. After #313 merged, I integrated the checkpoint with the reflection cursor and desire-review path so both inference actions share one guard and one crash-recovery state machine. Guard cleanup is exception-safe (b683044), and a new concurrency test proves overlapping desire reviews execute once. TypeScript and 13 sweep tests pass. No remaining blocking issue.

PR #315, opened 2026-07-25
Unify inference admission and lease lifecycle

oratis:
## Summary

- add one admission boundary for account-funded `/chat` and managed `/gw/*` inference
- apply the shared order: protocol validation → service/RPM guards → tenant lease → quota check → execution
- return an idempotently releasable permit that owns lease renewal for the full turn
- guarantee lease release on quota storage errors, upstream failures, non-stream responses, streaming disconnects, and normal completion
- remove the gateway's outer lease so malformed/oversized uploads cannot reserve a tenant before validation

## Why

Chat and the managed gateway had independently assembled the same billing controls in different places. That made ordering drift likely and split lease ownership across the router and handler. A single boundary gives future inference surfaces one contract and is the prerequisite for extracting a tenant runtime without duplicating billing semantics again.

## Behavior

Rejection payloads remain compatible (`service_paused`, `rate_limited`, quota errors, and `turn_in_progress`). Successful chat still receives its token budget hint; Mac/shared-token chat remains operator-funded and outside account admission.

## Validation

- `npm run typecheck` — passed after final edit
- `npm test` — 1388 passed, 1 skipped
- `npm run build` — passed
- focused tests cover limit short-circuiting, busy leases, quota denial cleanup, idempotent release/renewal, and storage-error lease cleanup

PR #316, opened 2026-07-25
Bound per-tenant runtime lifecycle

oratis:
## Summary

- replace the unbounded per-UID chat-context and prompt maps with a lifecycle registry
- create each tenant runtime single-flight under concurrent first requests
- pin runtimes while session/history/chat/reflection work is active
- evict idle runtimes by TTL and LRU with a hard tenant-count target
- keep prompt snapshots inside the same runtime so context and prompt memory are released together
- expose `LISA_TENANT_RUNTIME_TTL_MIN` (default 30) and `LISA_TENANT_RUNTIME_MAX` (default 100)

## Why

The cloud process retained every tenant's full loaded history and prompt forever. Memory therefore grew with lifetime tenant cardinality, and concurrent first requests could create duplicate contexts before either populated the old map. This establishes the TenantRuntime lifecycle seam before moving advisor/activity state and context compaction into it.

## Safety

Pinned runtimes are never evicted. The registry may temporarily exceed its target when every entry is active, then returns under the limit as leases release. Runtime creation failures clear their single-flight slot and remain retryable. The local/global runtime is unchanged and not subject to tenant eviction.

## Validation

- `npm test` — 1393 passed, 1 skipped
- `npm run build` — passed
- tests cover single-flight creation, pinned-entry safety, TTL expiry, LRU order, idempotent release, retry after failed creation, and configuration parsing

## Dependency

This is intentionally stacked on #315 because it consumes the unified inference permit lifecycle. Review and merge #315 first; this PR then becomes a focused TenantRuntime change.


oratis:
Code review: reviewed single-flight creation, pin accounting, TTL/LRU eviction, capacity behavior when every runtime is active, prompt-cache co-lifecycle, and release coverage across session/history/chat/reflect routes. Found an account-deletion race where an in-flight creation could reinsert a deleted runtime; fixed in 4e2ccaa with invalidation and a retry regression test. Integrated with latest billing/admission/body/autonomy main. TypeScript, build, and 30 focused tests pass. No remaining blocking issue.

PR #324, opened 2026-07-26
Add social connector host foundation

oratis:
## What changed

- adds the researched Sense social connector architecture and platform rollout plan
- defines and strictly validates `social-connector.json` manifests
- adds revisioned social drafts with immutable SHA-256 approval snapshots, expiry, and one-shot claim semantics
- surfaces connector manifests and draft state through `lisa sense social`
- preserves MCP tool annotations as metadata without treating untrusted hints as authorization

## Why

Conversation-driven publishing needs a host-enforced trust boundary before any platform connector can publish. Connecting an account must not silently authorize a post, and model-visible tools must not bypass the user final confirmation.

## Impact

This is the host foundation only; it does not publish to a live platform. Follow-up stacked PRs add the draft-only host tool/API/UI and the first Bluesky/Mastodon connectors.

## Validation

- `npm run typecheck`
- `node --import tsx --test src/mcp/client.test.ts src/sense/social/manifest.test.ts src/sense/social/drafts.test.ts`
- `git diff --check`

### DIFF
--- a/docs/PLAN_SENSE_SOCIAL_CONNECTORS_v1.0.md
+++ b/docs/PLAN_SENSE_SOCIAL_CONNECTORS_v1.0.md
@@ -1,0 +1,843 @@
+# PLAN · SENSE SOCIAL CONNECTORS — 对话式社交发布（v1.0）
+
+> 调研与设计快照：2026-07-26  
+> 状态：**方案锁定，S0 host foundation 已开始落地**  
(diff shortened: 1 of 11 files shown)


oratis:
Post-merge review finding:

`mutate()` currently sorts every draft and applies `.slice(-MAX_DRAFTS)`. That can evict a non-terminal draft (`approved`/`publishing`) when enough newer drafts exist. A connector may then publish successfully while `completeSocialDraftPublish()` can no longer find the draft to persist receipts, leaving an externally visible post with no durable host outcome.

I am addressing this in a dedicated follow-up: retain every non-terminal draft, prune only terminal history, and reject new drafts once the active-draft ceiling is reached.

PR #325, opened 2026-07-26
Sense: add trusted social draft workflow

oratis:
## What
- add a model-visible `social_compose` tool for connector discovery, drafting, revision, and confirmation requests
- keep approval out of the model tool and require a loopback or authenticated user session
- add authenticated Sense social APIs with no public publish route
- render connected-media status, full immutable approval snapshots, approve, and cancel controls in Sense
- persist approval expiry and block social composition in autonomous runs

## Why
This is the host enforcement layer for conversational social publishing. It makes the exact post snapshot reviewable while ensuring the model cannot approve or publish its own work.

## Impact
Draft and approval state is stored under the user-scoped LISA home with mode 0600. No platform is contacted by this PR, and no social publish endpoint exists. This PR is stacked on #324.

## Checks
- `npm run typecheck`
- `npm test` (1456 passed, 1 skipped)
- `git diff --check`

### DIFF
--- a/src/sense/social/drafts.test.ts
+++ b/src/sense/social/drafts.test.ts
@@ -5,6 +5,7 @@ import path from "node:path";
 import { afterEach, beforeEach, describe, test } from "node:test";
 import {
   approveSocialDraft,
+  cancelSocialDraft,
   canonicalJson,
   claimApprovedSocialDraft,
   createSocialDraft,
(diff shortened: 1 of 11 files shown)


oratis:
Post-merge review finding:

`GET /api/sense/social/connectors` returns discovery objects directly. Those objects include `manifestPath`, which is an absolute local filesystem path. The Sense UI does not need it, and returning it exposes the user/plugin directory layout to authenticated remote clients.

I am addressing this in a dedicated follow-up by projecting connector responses onto an explicit public shape (`plugin`, validated manifest or error, and redacted accounts) and adding a regression test that the local path never appears in the response.

PR #326, opened 2026-07-26
Sense: ship Bluesky and Mastodon connectors

oratis:
## What
- add installable first-party LISA connector plugins and procedural Skills for Bluesky and Mastodon
- link Bluesky through app-password session exchange and Mastodon through a verified user token
- publish text, links, images, and video through official APIs with runtime validation and idempotency
- add a private host media store with MIME sniffing, integrity hashes, and image metadata stripping
- add the deterministic approved-draft runner, per-target receipts, structural audit, and pause switch
- remove manifest-bound publish/disconnect tools from every model-visible toolset
- add CLI install/connect/media commands and MCP stdio handshake coverage

## Why
This is the first live connector slice after the host safety layer. It proves the Connector + Skill contract on an open network and a federated platform without letting the model acquire publish authority.

## Impact
Users can install both connectors with `lisa sense social install`, link accounts in a secret-safe TTY, prepare posts in chat, approve the exact Sense snapshot, and publish through the hidden host runner. The connector account/token store and audit files are private (0600). This PR is stacked on #325.

## Checks
- `npm run typecheck`
- `npm test` (1466 passed, 1 skipped)
- real MCP stdio initialization/list-tools/close test
- `git diff --check`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -471,6 +471,16 @@ async function main(): Promise<void> {
     }
   }
   const mcpTools = mcpConnections.flatMap((c) => c.tools);
+  // Connector publish/disconnect operations remain available to the trusted
+  // host runner but are never placed in the model-visible toolset.
+  const { discoverSocialConnectors, hiddenSocialMcpToolNames } = await import(
+    "./sense/social/manifest.js"
+  );
+  const socialConnectors = await discoverSocialConnectors();
+  const hiddenSocialToolNames = hiddenSocialMcpToolNames(socialConnectors);
+  const modelMcpTools = mcpTools.filter(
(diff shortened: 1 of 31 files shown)


oratis:
Post-merge review finding:

The Mastodon path forwards the host idempotency key, but `publishBluesky()` ignores it: it calls `com.atproto.repo.createRecord` without an `rkey` and generates a fresh `createdAt` on every attempt. A retry after an unknown outcome can therefore create a duplicate public post even though the MCP tool advertises `idempotentHint: true`.

I am addressing this in a dedicated follow-up by deriving a deterministic AT Protocol record key from the host idempotency key, using a stable approval timestamp, publishing through `putRecord`, and testing that repeated identical inputs target the same record with the same body.

PR #357, opened 2026-08-13
feat(sessions): 系统提示词入会话日志 + 离线重放（H3）

oratis:
> 计划见 #356 的 `docs/PLAN_HARNESS_ALIGNMENT_v1.0.md` §4（H3）。本 PR 是其 Step 1。

## 问题

会话 JSONL 一直记录 `tool_use` / `tool_result`（它们是 `StoredMessage` 的 content block），这点比预想的好。真正缺的是**系统提示词**：

- 它由 soul + memory + skills + mood + KB 摘要动态构建（`prompt.ts:66`）；
- 而且会话**中途会热重载** —— `soul_patch` / `memory` 写入后下一轮就换了一份（`agent.ts` 的 `hotReload`），只发一个 `system_prompt_rebuilt` 事件，不落盘；
- 结果：**拿一条历史会话，重建不出当时模型看到的 persona**。

ROADMAP 1.0 判定标准第 4 条要求 Reve 产出可复现的 drift / coherence 指标，而 drift 的定义就是"她的自我描述随时间怎么变"。日志里没有当时的自我描述，这个指标只能在线测一次 —— 无法离线复算、无法换个定义重算、无法做消融。

## 改动

| | |
|---|---|
| 会话格式 | v1 → **v2**，新增 `prompt` 条目 `{ts, fingerprint, text, reason}` |
| 去重 | `fingerprint` 是正文内容哈希，连续相同不重复写 → "第 N 条时生效的提示词" = 最近的前一条 prompt 条目；从不自我修改的长对话只花一条 |
| 触发点 | `runAgent` 新增 `onPromptPersist`，在**每次 provider 调用之前**触发 —— 即提示词真正对模型可见的那一刻。写在调用前，所以中途崩溃的轮次也留得下"当时问的是什么" |
| 接线 | 三个持有 `SessionStore` 的表层全接：REPL / web / 渠道。子代理与 managed agent 无会话，按设计不接 |
| 交付物 | `src/sessions/replay.ts` 离线重建每轮 `(systemPrompt, messages)`；`scripts/replay-session.ts` CLI 包装 |

失败处理：持久化异常只写一行日志，绝不拖垮活轮次。

## 诚实边界

写进了 `replay.ts` 的模块头注释，也在这里说清：

- `systemPrompt` 是**逐字精确**的（CLI 与 web 都持久化了含 suffix 的最终串）；
[...]

PR #359, opened 2026-08-13
feat(sandbox): 三档模式 + fail-closed + 会话级固定（H2）

oratis:
> **栈式 PR：base 是 #358（H1），不是 main。** 计划见 #356 的 `docs/PLAN_HARNESS_ALIGNMENT_v1.0.md` §3。
> 合并 #358 时**请勿加 `--delete-branch`** —— 会连带关掉本 PR。

## 三个洞，按严重度

**1. `write` / `edit` / `apply_patch` 此前完全不受沙箱约束。**

`wrapForSandbox` 唯一调用点是 bash 工具，而三个写工具都是 `path.resolve(ctx.cwd, input.path)`，绝对路径与 `../` 一律照收。也就是说 `LISA_SANDBOX=1` 时 **bash 被关进 cwd、write 能写满盘** —— 用户拿到了一个自己并不拥有的保证。

**2. 非 macOS 静默降级**成不受限 `/bin/bash -lc`。现在无法强制时抛 `SANDBOX_UNAVAILABLE` 并拒绝执行。"以为沙箱开着其实没开"比"知道没有沙箱"更危险。

**3. 权限没有会话级固定** —— 改设置会影响正在跑的任务。

## 三档（沿用 dsh 的词汇，不自造）

| mode | 文件读 | 文件写 | shell |
|---|---|---|---|
| `read-only` | 任意 | 拒绝 | 无可写路径 |
| `workspace-write` | 任意 | 工作区 + 临时目录 | 工作区 + 临时目录 |
| `danger-full-access`（默认） | 任意 | 任意 | 不受限 |

两个刻意的决定（都写进了代码注释，不是疏漏）：

- **读在所有模式下都不设界。** Seatbelt 无条件放行 `file-read*`；若只把 fs 的读收紧，就重新制造了 H2 要消灭的那种不对称，还会弄坏一切读工作区之外（含 `~/.lisa`）的工具。"能不能读"由上一层的工具子集决定。
- **临时目录在 `workspace-write` 下默认可写。** 同一模式下 bash 本来就能写 `/tmp`，再去拦 `write` 是表演不是边界。

## 默认值不变

默认仍是 `danger-full-access`：本地有人值守的 REPL 与往 shell 里敲命令是同一信任姿态，**静默收紧所有人的现有配置不是本次改动的职责**。本次修的是"把沙箱打开时它真的管住文件写"。

按 surface 收紧无人值守默认值（计划 §3 的表：dispatch→`workspace-write`、idle/渠道→`read-only`）会改行为，**另开 PR**。

## 实现

落在 #358 的 seam 上：sandboxed 提供方包住 local 提供方，`capsOf` 按解析出的模式选世界 —— 所以 `LISA_SANDBOX=1` 现在自动覆盖全部 fs/shell 工具而不只是 bash。符号链接逃逸用"最近存在祖先的 realpath"处理（纯路径检查会放过它）。

## 测试（27 例）

`../` 逃逸 / 绝对路径逃逸 / **符号链接穿透** / `edit` 与 `apply_patch` 的 update 与 delete / `read-only` 连工作区内也拒 / 读不受影响 / 临时目录默认可写 / 不支持平台 fail-closed / Seatbelt 策略逐档断言 / 模式解析优先级 / 拼写错误报错而非静默回退 / 会话级固定不被后续设置变更加宽。

**本机实测**（`LISA_SANDBOX=1`，目标 `~/lisa-h2-probe.txt`）：

```
bash  → touch: Operation not permitted   (Seatbelt)
write → 拒绝：sandbox mode "workspace-write" confines writes to <workspace>
```

此前第二行会是 `ALLOWED`。全量 **1592 通过 / 0 失败**。

## 与 #357 的交互

两者都改 `SessionHeader`：#357 加 `version: 2`，本 PR 加可选的 `sandboxMode`。互不冲突（可选字段无需版本位），但 `types.ts` 会有一处文本冲突，取并集即可。

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,7 +462,7 @@ On Linux, `lisa heartbeat install` prints a cron line for you to add to `crontab
 | Tool | Purpose |
 |---|---|
 | `read` `write` `edit` `apply_patch` | File ops (single + batched) |
-| `bash` | Shell (with optional macOS Seatbelt sandbox via `LISA_SANDBOX=1`) |
+| `bash` | Shell (confined by the sandbox mode — see below) |
 | `grep` `ls` | Search + listing |
 | `task` | Spawn a focused sub-agent in its own context window |
 | `dispatch_agent` `signal_agent` `dispatch_status` | Launch / stop / track agents she runs (managed + PTY); refuses directories another agent owns |
@@ -476,6 +476,20 @@ On Linux, `lisa heartbeat install` prints a cron line for you to add to `crontab
 | `memory` `memory_search` | Memory CRUD + TF-IDF search across all past sessions |
 | `kb_search` `kb_read` `kb_list` `kb_links` `kb_add` `kb_write` `kb_ingest` | Personal knowledge base — search + read/list, explore the link graph, add a source, write/maintain a wiki page, ingest a URL (WeChat / Bilibili / YouTube / any article) |
 | `set_mood` | Switch her visible portrait to one of 114 moods |
+
+### Sandbox modes
+
+File and shell tools share one **execution world**, so they can never be bounded to different roots:
+
+| mode | file reads | file writes | shell |
+|---|---|---|---|
+| `read-only` | anywhere | refused | no writable path |
+| `workspace-write` | anywhere | workspace + temp | workspace + temp |
+| `danger-full-access` *(default)* | anywhere | anywhere | unconfined |
+
+Set with `LISA_SANDBOX_MODE`; `LISA_SANDBOX=1` remains an alias for `workspace-write`. Shell confinement is enforced by the OS — macOS Seatbelt, Linux bubblewrap. **On a platform where the requested mode cannot be enforced, the command is refused (`SANDBOX_UNAVAILABLE`) rather than silently run unconfined** — believing a sandbox is on when it is off is worse than knowing there is none. File confinement is enforced in-process, so it holds everywhere, and it resolves symlinks so a link inside the workspace cannot be written through.
+
+The mode is fixed when a session is created and recorded in its header: changing the setting never widens what an already-running task may do.
 | `soul_patch` `soul_journal` `soul_feel` `soul_read` | Her soul-editing tools (hers alone) |
 | `soul_history` `soul_diff` | Read the git-backed history of her own soul (every change committed with attribution) |
 | `soul_object` | Architectural objection — flags a constitutional concern; the agent loop forces it to be surfaced in her reply |
@@ -551,7 +565,8 @@ LISA_PTY_CLAUDE_CMD=claude            # override the `claude` binary path
 LISA_PTY_CODEX_CMD=codex             # override the `codex` binary path
 
 # Sandbox
-LISA_SANDBOX=1                        # opt-in macOS Seatbelt for `bash`
+LISA_SANDBOX_MODE=workspace-write     # read-only | workspace-write | danger-full-access
+LISA_SANDBOX=1                        # legacy alias for workspace-write
 LISA_SANDBOX_NETWORK=0                # block network in sandbox
 
 # Web
@@ -691,7 +706,8 @@ src/
 ├── kb/                     ★ personal knowledge base (Karpathy 3-layer: sources + wiki) + kb_* tools + TF-IDF
 ├── mail/                   read-only IMAP / Gmail-OAuth mailbox — classify + daily digest + alerts
 ├── sessions/               JSONL store + list + resume + paginated read
-├── sandbox/                macOS sandbox-exec policy + wrapper
+├── capabilities/           fs / shell capability seam — local, sandboxed, in-memory providers
+├── sandbox/                sandbox modes + macOS Seatbelt / Linux bwrap policies
 ├── mcp/                    config + stdio client (wraps MCP tools as Lisa tools)
 ├── plugins/                claude-code-style plugin loader
 ├── hooks/                  PreToolUse / PostToolUse / SessionStart / etc.
--- a/src/capabilities/index.ts
+++ b/src/capabilities/index.ts
@@ -45,7 +45,11 @@ export function defaultCapabilitiesFor(
 }
 
 export function capsOf(ctx: ToolContext): Capabilities {
-  return ctx.caps ?? defaultCapabilitiesFor(ctx.cwd);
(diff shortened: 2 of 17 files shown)


oratis:
## Security review — 2 HIGH ship-blockers; recommend NOT merging until enforcement is wired

The fail-closed *plumbing* here is genuinely well done and should be kept: `wrapForSandbox` throws `SandboxUnavailableError` on an unsupported platform / missing bwrap and the sandboxed provider does **not** catch it (no silent unconfined fallback); the symlink write-escape is really blocked (`realpathOfNearestExisting` + `atomicWrite` renames over a planted link); `resolveSandboxMode` throws on a typo rather than defaulting; and `read-only` Seatbelt grants no writable path but `/dev/null`. But two HIGH issues make the PR's headline guarantee false, so I've **held this PR** (merged the rest of the batch — H1 #358, H3 #357, P1 #360, plan #356) rather than ship a sandbox that doesn't confine.

### HIGH-1 — session pinning (会话级固定) is recorded but never enforced
`header.sandboxMode` has exactly one producer (`sessions/store.ts:48`) and **zero enforcement consumers** — `grep sandboxMode src/` outside tests/types confirms it. Every tool call resolves its world through `capsOf(ctx) = ctx.caps ?? defaultCapabilitiesFor(ctx.cwd)` with **no mode**, and `ctx.caps` is never assigned in non-test `src/` — so `defaultSandboxSpec` re-reads `process.env.LISA_SANDBOX_MODE` **fresh on every call**. Consequences:
- An explicit `SessionStore.create({ sandboxMode: "read-only" })` pin does nothing — the next `write`/`bash` re-derives `danger-full-access` from env and runs unconfined. The store test only proves the header *string* round-trips, not that enforcement honors it.
- The README's "fixed when a session is created … changing the setting never widens what an already-running task may do" is false: flip the env mid-flight and the next tool call uses the new mode.
- Because the mode is a process-global env var, **concurrent sessions can't have different modes** — the exact opposite of "会话级固定."

**Fix:** thread the pinned mode into the context that reaches tools — carry `sandboxMode` on `ToolContext`, have `capsOf` use `defaultCapabilitiesFor(ctx.cwd, ctx.sandboxMode)`, and set it from `session.header.sandboxMode` at the turn-build sites (`web/server.ts:3971`, `cli.ts:718`, `subagent.ts:47`, `agents/managed.ts:170`, `channels/router.ts:127`). Add a test that a `read-only`-pinned session actually **rejects a write through `writeTool`**.

### HIGH-2 — default mode is `danger-full-access`, including on untrusted-input surfaces
`sandbox/mode.ts:71` defaults to fully-unconfined, and (given HIGH-1) that's what everything runs under. The surfaces processing the **least-trusted** input — `channels/router.ts` (inbound Slack/DM), idle/heartbeat runners, mail + KB-feed classification — create contexts with no mode, so a prompt injection from an inbound message or a fetched RSS/WeChat/YouTube page drives `write`/`bash` with zero confinement.

**Fix:** default the unattended/untrusted surfaces to at least `workspace-write` (pass an explicit `sandboxMode` when those sessions/contexts are created), keeping the interactive `danger-full-access` default for the local user. (Depends on HIGH-1 so the pin is honored.)

### MED-3 — `shell.exec` bypasses the sandbox in every bounded mode
`capabilities/sandboxed.ts:150` wraps only `run`; `exec` is `localShell.exec` (unconfined). `grep` uses `exec`, so in `read-only`/`workspace-write` it runs with no OS confinement — harmless today (grep only reads) but it's the "one call site forgot the wrapper" invariant-break H2 exists to eliminate. Route `exec` through the same confinement as `run`.

### Recommendation
The gap is **wiring, not architecture** — the sandbox providers, fail-closed, and symlink defenses are solid. But the wiring is cross-cutting and security-critical (6 context-build sites + defaulting the untrusted surfaces + a real enforcement test), and it should be done with full attention, not squeezed in. I recommend landing it as a focused follow-up PR on top of the now-merged H1. Holding #359 open (base still on the H1 branch) until then.

oratis:
## Fix applied — the pin now enforces (rebased onto current `main`)

Per the decision to wire it rather than hold, the **core ship-blocker (HIGH-1) is fixed and proven**: the sandbox now actually enforces a pinned mode instead of re-reading `process.env` every call. typecheck clean; full suite **1630 pass / 0 fail**.

**What changed**
- `ToolContext.sandboxMode` is now a live field, and `capsOf` resolves the world from it — `ctx.caps ?? defaultCapabilitiesFor(ctx.cwd, ctx.sandboxMode)` — falling back to the env default only when nothing is pinned. This is the single point where a pin takes effect.
- **Enforcement test** (`seam.test.ts`): a `read-only` pin refuses a write **even when `LISA_SANDBOX_MODE=danger-full-access`** — the pin wins over the env, and a bounded pin selects a different (sandboxed) fs than the unconfined default. This is exactly the guarantee that was silently false before.
- **CLI** threads `session.header.sandboxMode`, so the mode is **frozen at session creation** and can't be widened mid-session by an env change.
- **Subagent** + **channel router** carry `sandboxMode`, so a subagent can't escape its caller's confinement and an operator exposing fs/shell to a channel (`unsafeFullTools`) can pin a bounded mode.

**Deliberately deferred (documented, so this doesn't ship a false promise):**
- **HIGH-2 default-flip** — defaulting the unattended/untrusted surfaces (idle, heartbeat, feed/mail classification) to `workspace-write`. The plumbing is now in place, but bounded modes fail **closed** without an OS sandbox (bwrap on Linux / Seatbelt on macOS), so flipping the default blind could break autonomy/channels on a host without bwrap. Should land once the deployment's sandbox support is confirmed. The env-level `LISA_SANDBOX_MODE` already confines everything globally in the meantime.
- **Web-turn threading** (`server.ts` — thread `chat.session.header.sandboxMode`) and **task.ts inheritance** (its `execute` has no `ctx`; needs `deps` to carry the mode).
- **MED-3** — routing `shell.exec` through the same confinement as `run`. Harmless today (only `grep` uses `exec`, and reads are unbounded in every mode by design); the correct fix needs an argv-aware `wrapForSandbox`.

Net: the sandbox genuinely enforces a set/pinned mode now (tested), which resolves the "recorded but never enforced" ship-blocker. The remaining items are safe, well-scoped follow-ups rather than false-security.

PR #365, opened 2026-08-21
fix: verified findings from the analytics-plan research pass (consent boundary, soul fingerprint, dispatch exit status, doc counts)

oratis:
## What this changes

While building the telemetry/analytics plan (`docs/analytics-plan-2026-08-21.md`, included here), the research pass had to establish a lot of ground truth about this repo, and it turned up a set of real defects along the way. Each one was then independently verified against the source before being fixed. This PR is that batch: one security fix at the cloud HTTP boundary, one privacy fix in the soul birth path, two correctness fixes in the dispatch ledger, one dead-code removal, three documentation corrections, and a new CI guard for the published "no telemetry" promise. Every finding has a regression test, and each fix is its own commit so anything contentious can be dropped independently.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] New channel adapter
- [ ] New mood / pixel asset
- [ ] Plugin example / MCP example
- [x] Docs only *(partly — three of the nine commits)*
- [x] Refactor / chore

---

## ⚠️ Read this before merging

Three commits change behavior or need a human decision. Everything else is inert.

**1. `fix(soul): keep the device fingerprint out of the birth prompt` — touches `src/soul/*`, so per CONTRIBUTING it wants a discussion issue first. I could not open one; please treat this as the proposal.**

`seed.bornOn` is `sha256(hostname + username)` — unsalted, over two low-entropy and often-guessable strings, stable across rebirths on the same machine. `dreamSoul()` serialized the **whole** seed into the birth prompt, so that fingerprint was sent to whichever model provider is configured, on every birth. It contributes nothing to the dream (personality comes from `randomness` + `bigFive`), so this was pure unnecessary egress.

The behavior change: the birth prompt is now one field shorter, so a **new** birth will dream a slightly different soul than it would have before. Already-born souls are untouched. Stored seeds still contain `bornOn`; only what crosses the wire narrowed.

**What this does NOT fix:** `GET /api/soul` still serves the full seed, `bornOn` included, to authenticated clients — the iOS companion reads that endpoint. Narrowing that is a separate decision, flagged in the type comment and in the follow-ups below.

**2. `fix(dispatch): capture exit status…` changes `dispatch_status` output.** `✓ finished` is gone. Labels are now `▶ running` / `✓ exit 0` / `✗ exit N` / `✗ killed by SIG…` / `• exited (status not captured)`. The last one is the honest case for a dispatch that outlived LISA's own process — the child is detached and `unref`'d, so the exit listener only fires while LISA is alive (true for `lisa serve` and a live REPL, false for a one-shot CLI invocation). Anything parsing that string will need updating.

**3. `refactor(advisor): drop the repeated_failure category`** removes a member from the exported `SuggestionCategory` union and the `errorCommandCounts` field from `AdvisorState`. Existing `advisor-state.json` files are unaffected — `loadAdvisorState()` spreads parsed JSON over `emptyAdvisorState()`, so a leftover key is ignored.

---

## What's fixed

| # | Severity | Fix |
|---|---|---|
| 1 | **security** | **`/api/consent` routes were missing from the cloud deny-list.** `src/consent/store.ts` resolves `~/.lisa/consent.json` with its own `lisaHome()` instead of the per-user home scope in `src/paths.ts`, so consent state is per-machine, not per-tenant. Every other host-control route was denied in the hosted edition; these were not. Any signed-in tenant could read every tenant's grant list, grant `mail` deployment-wide, or `POST /api/consent/revoke-all` and stop the mail digest sweep for everyone. Added the prefix; the existing matcher already covers the bare root and sub-paths without over-matching siblings like `/api/plans-public`. |
| 2 | **security** | **Device fingerprint sent to the model provider on every birth.** See the note above. |
| 3 | correctness | **Dispatch never captured an exit code.** `launchAgent()` only attached an `error` listener for 150 ms, which catches launch failures and says nothing about the run, so a crashed agent and a clean one both rendered as `✓ finished`. Now records `exitCode` / `exitSignal` / `exitedAt`. Subtlety worth noting: the listener has to be attached **before** the 150 ms launch race — an agent that exits inside that window emits `close` first, and a listener attached afterwards never fires. Both orderings are covered by tests. |
| 4 | correctness | **Pid reuse could make `signal_agent` kill an unrelated process group.** Entries are retained 24 h and matched by bare pid; `isAlive()` was just `kill(pid, 0)`, so a recycled pid answers "alive" and `signalGroup()` then sends SIGTERM/SIGKILL to `-pid`. That contradicts the invariant in `signal_agent.ts`'s own header ("LISA cannot kill an arbitrary process"). Now records a start-time fingerprint (`/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` elsewhere) and requires it to match. A failed probe means "cannot tell" and is treated as a match; entries without a token keep the old behavior, so nothing silently disappears. |
| 5 | hygiene | **`repeated_failure` was a declared advisor category no detector ever emitted**, with an equally inert `errorCommandCounts` under a comment describing a rolling memory that never rolled. `docs/PRODUCT_REVIEW_v0.9.md` already flagged both. Deleted. |
| 6 | docs | **Observer count was wrong in both READMEs.** They said "all five observers"; ten ship. The README also contradicted itself — the directory tree 490 lines below already listed seven plus an ellipsis. Now says ten, split into the five coding-agent adapters and the five others, and adds the fact the old sentence was likely reaching for: only three are enabled by default. |
| 7 | docs | **Advisor categories in the READMEs listed 4 of 5.** Added `cost_spike`. `repeated_failure` deliberately not documented — it is deleted in this PR, and documenting a card the product never produces would make the docs more wrong, not less. |
| 8 | docs | **The PR template pointed regression tests at a `test/` directory that does not exist.** `npm test` globs `src/**/*.test.ts`, so a test written where the template said would never run and CI would go green having never executed it. CONTRIBUTING.md already had it right. |
| 9 | prevention | **New guard for the published "no telemetry" promise.** See below. |

## The no-telemetry guard

The promise is currently true — no analytics SDK, no tracking pixel, no third-party script, no phone-home anywhere in the tree. Nothing enforced it, so one `npm install` of a convenience wrapper or one snippet pasted into a layout would turn a published privacy claim into a false statement with nobody noticing. `src/no-telemetry.test.ts` checks a deny-list over `src/`, `website/src/`, the iOS companion sources, and `package.json` dependency names. It runs in the normal suite, so it gates every PR through `ci.yml` and blocks `prepublishOnly` — no CI edit needed.

Two deliberate choices:

- **Promise pages are listed by path, not found by grepping English phrases.** The Chinese pages say 无遥测 and 没有分析 SDK, which no English keyword search would match — a keyword-driven guard would silently have covered only half the site. Listing paths also means renaming a promise page fails loudly instead of shrinking coverage.
- **Tokens are SDK-shaped, not bare words.** `segment`, `heap` and `plausible` appear in ordinary prose and identifiers here (13, 35 and 2 files), and `amplitude` is an audio term. A guard that cries wolf gets deleted. Two tests pin this from both sides: one asserts the list really matches `posthog-js` / `googletagmanager` / `@sentry`, so it can never go vacuous, and one asserts it stays quiet on the English words.

Verified by injecting a tracker `<script>` into the Chinese homepage and confirming three of these tests fail, then reverting.

## Before / after

```
# dispatch_status, agent crashed with exit 1
- ✓ finished · claude (pid 4823, 2m ago) · id 4823-mt2f
+ ✗ exit 1 · claude (pid 4823, 2m ago) · id 4823-mt2f

# dispatch_status, agent outlived LISA — status genuinely unknown
- ✓ finished · codex (pid 5010, 1h ago) · id 5010-mt2g
+ • exited (status not captured) · codex (pid 5010, 1h ago) · id 5010-mt2g
+   note: the agent outlived this LISA process, so its exit status was never seen …
```

```diff
# src/soul/birth.ts — what the model provider receives
- `Seed:\n${JSON.stringify(seed, null, 2)}\n\nBirth yourself.`
+ `Seed:\n${JSON.stringify(seedForPrompt(seed), null, 2)}\n\nBirth yourself.`
```

## Follow-ups for a human (not fixable in this PR)

1. **Fix the underlying cross-tenant consent scoping.** This PR only closes the HTTP route. `src/consent/store.ts` still writes one shared `~/.lisa/consent.json` outside the per-user home scope, so consent remains per-machine in the hosted edition. It is also a hard prerequisite for hanging telemetry consent off that framework (plan §9.2).
2. **Decide what `GET /api/soul` should expose.** It still serves the full seed including `bornOn` to authenticated clients. Narrowing it is a breaking change for the iOS companion, so it needs an owner's call.
3. **Decide whether `bornOn` should exist at all.** Its only read is a display string (`seed.bornOn.slice(0, 8)`). `crypto.randomBytes` would serve that with no linkability — but it changes stored soul data.
4. **Open the discussion issue CONTRIBUTING asks for on the `src/soul/*` change**, or drop commit `eb92134` from this PR.
5. **Seven other modules still define a private `lisaHome()`** that bypasses `homeScope` (`sense/log.ts`, `dispatch-ledger.ts`, `control/policy.ts`, `web/push.ts`, `mail/{store,accounts}.ts`, `takoapi/ledger.ts`). Currently unreachable in cloud mode because both the route deny-list and the tool allow-list hold — but that is two gates deep, not a fix. Do **not** bulk-change `web/{accounts,devices,otp,sessions-auth}.ts`; those four are correctly process-global.
6. **The plan doc references an absolute local path** (`/Users/oratis/Documents/Claude/analytics-blueprint.md`) as its methodology source. Harmless but you may want it scrubbed before this lands on a public repo.

## Checklist

- [x] `npm run typecheck` passes
- [x] `npm run build` passes
- [x] Regression test co-located with the source (`src/**/*.test.ts`) — added for all seven code findings
- [ ] **`src/soul/*` change is NOT linked to a discussion issue** — see the warning above; this is the one checklist item this PR knowingly does not satisfy
- [x] Commit messages are lowercase, imperative, no emoji
- [x] No `any` introduced

Full gate run locally, all four CI steps: `check:api-contract` clean · `typecheck` clean · `npm test` **1674 pass / 0 fail / 1 skipped (1638 → 1675 tests)** · `build` clean, 0 `*.test.js` leaked into `dist/`.

## Related issues

n/a — findings came from a verification pass, not a filed issue.


oratis:
## Review pass — four fixes pushed

A high-effort review of this branch (10 finder angles, adversarial verification, gap sweep) surfaced 15 findings. Four are addressed in the four commits just pushed; the rest are listed below as follow-ups.

Every fix here carries a test that was checked by reverting the fix and confirming the suite goes red — the lesson from finding #4 below.

### Fixed here

**1. `/api/push/*` was left off the cloud deny-list** — the same defect class as the `/api/consent` entry this PR adds, one file over. `src/web/push.ts` resolved `~/.lisa` with its own private `lisaHome()`, outside the per-uid `homeScope`, and `PushSubscription` has no owner field. The hosted edition runs one container with one `LISA_HOME` and separates tenants only by `homeScope.enterWith(homeForUid(uid))` per request, so `push.json` was one shared file and none of the five routes sat behind `denyRemote`, a loopback check, or an owner check. Verified by running the real modules under two uid scopes: tenant B read tenant A's subscription and could unregister it — exposing ntfy topics, which are themselves the send/read secret for that channel. The store stays machine-wide on purpose (every `PushBridge` producer is host-level), so the fix is to deny the routes in cloud and say `lisaGlobalHome()` explicitly instead of re-deriving `~/.lisa`.

**2. The deny-list is bypassable by path.** `isCloudDeniedRoute` matches the *normalized* pathname while every handler in `server.ts` matches the *raw* `req.url`. `GET /api/agents/recap/%2e%2e/%2e%2e/%2e%2e` normalizes to `/` — not denied — while still satisfying `url.startsWith("/api/agents/recap")`, so the handler runs. Same shape reaches `/api/agents/steps`, `/api/agents/transcript`, `/api/agents/pty/*/output`, `/api/dispatch/status` (raw captured agent stdout) and `/api/mail/accounts`; `//api/consent/grant` reparses its authority, so the entry added in this PR is itself bypassable. Verified against the real exported function. Rather than teach ~80 route checks to normalize, non-canonical paths are now rejected before anything routes — fails closed for routes added later.

**3. `entryIsAlive` ignored the exit this PR just taught the ledger to record.** After `recordExit(id, 3, null)`, `entryIsAlive()` returned true, `statusLabel()` returned `▶ running`, and `findDispatch()` still returned the entry — with `exitCode: 3` in the same JSON object. It matters beyond cosmetics because `recordDispatch` runs *after* the 150 ms launch race, so a fast-crashing agent is already reaped when `processStartToken` shells out and gets no `startToken` at all; such entries fall back to bare pid identity for 24 h, and `signal_agent cancel` would signal whatever process group later owns that pid. Gated on `exitedAt`, not `exitCode` (a signal death legitimately stores `exitCode: null`).

**4. The soul privacy fix had no test that could fail.** All three new tests assert on `seedForPrompt`'s own output; one re-derives the payload under a comment calling it "the actual wire shape". Reverting `birth.ts` to `JSON.stringify(seed, null, 2)` left the suite at 9 pass / 0 fail — the headline privacy fix of this PR was silently revertible. `dreamSoul` already takes its provider as a parameter, so it is now exported and driven with a capturing fake.

### Follow-ups (not addressed here)

- `ps -o lstart=` is TZ- and locale-dependent: measured four different tokens for one unchanged pid across `TZ`/`LC_ALL` settings. A `lisa serve` under launchd and a `lisa` CLI run from a configured shell disagree, so every running dispatch reports dead and a runaway agent becomes uncancellable. Pass `LC_ALL=C TZ=UTC` to the `spawnSync`, or normalize to epoch seconds.
- `startToken` is probed after the 150 ms race, so the pid-reuse guard is absent for exactly the short-lived dispatches most at risk. The pid is known synchronously after `spawn()`.
- The Linux `lt:<starttime>` token is not boot-scoped, so it can collide across a reboot inside the 24 h window.
- `spawnSync("ps")` measured 1.79 ms vs 0.001 ms for `kill(pid,0)`; `listLiveDispatches` probes each entry twice and `GET /api/dispatch/list` maps it over an unbounded ledger inside the request handler, blocking the event loop. Line 198's `entryIsAlive(e) || e.startedAt >= cutoff` is also ordered expensive-first.
- `toDispatchView` was not extended with the exit fields, so the web dashboard renders `exit 127` as a green "Done" and the iOS ledger as "Alive: no" — the fix lands on 1 of 3 surfaces.
- The no-telemetry guard has three holes, each confirmed by planting a canary and watching the suite stay green: `SKIP_DIRS` excludes `assets`/`build` at any depth (so `website/public/assets`, `src/web/assets` — which ships to npm — and `packaging/build` are unscanned); `website/package.json` and `website/astro.config.mjs` are never read, which is the standard way to add analytics to an Astro site; and the external-script test walks only `website/src`, not `website/public`.
- `src/integrations/hub.test.ts` permanently empties the built-in registry for the rest of the file (ESM module cache means the `import`-driven `registerIntegration()` side effects never re-run after `_resetIntegrationsForTest`). It passes today only by being the last `describe`; the next test appended after it fails with `unknown integration "claude-code"`.
- `saveLedger` is a bare `writeFileSync` and `recordExit` adds a second concurrent read-modify-write from an async listener. `atomicWrite()` already exists at `src/fs-utils.ts:27`.
- README/README.zh-CN: widening "all five observers" to "all ten" dragged the parenthetical "every adapter has a privacy test" over five adapters it was never written for — `managed` and `pty` have no test file at all.
- `loadOrchestratorConfig` does `parsed.integrations ?? DEFAULT…` (whole-map replace, not per-key merge), so a partial `~/.lisa/agents.json` silently drops the three default observers. Pre-existing, but this PR adds the README's only pointer to that file.

### Still needs your call

The three items in the PR description remain open: the `src/soul/*` discussion issue that CONTRIBUTING asks for, the `dispatch_status` output-string change, and the `SuggestionCategory` union member removal.

oratis:
## 第二轮评审 — 六条 follow-up 已修，带回退验证

上一轮留了 10 条 follow-up。这轮实证复核了每一条，修掉其中六条（四个提交），并核实了两处此前没被质疑的新增逻辑。

### 已修

**1. `ps -o lstart=` 不是进程身份，而是本地时钟的一种渲染。** 这是 follow-up 榜首，也是唯一一条会导致**失控 agent 永久无法取消**的。实测同一个不变的 pid：

| 环境 | 令牌 |
| --- | --- |
| `TZ=UTC` | `Mon Sep  7 13:23:46 2026` |
| `TZ=Asia/Tokyo` | `Mon Sep  7 22:23:46 2026` |
| `TZ=America/New_York` | `Mon Sep  7 09:23:46 2026` |
| `LC_ALL=de_DE.UTF-8` | `Mo.  7 Sep. 21:19:50 2026` |
| `LC_ALL=zh_CN.UTF-8` | `一  9月/ 7 21:19:50 2026` |

launchd 下的 `lisa serve` 与登录 shell 里的 `lisa` CLI 因此对每一个令牌都不一致 → `isAlive()` 判所有运行中的 dispatch 为死 → `signal_agent` 走进 "had already exited" 分支，**把账本行删掉而不是发信号**。失控 agent 从此不可取消，且再也找不回来。探测环境已钉死 `LC_ALL=C / TZ=UTC`。

顺带把令牌方案版本化（`ps1:` / `lt1:`），只在同方案内比较。否则升级瞬间，所有用旧 locale 写入的存量条目会集体误判为死亡 —— 与失败探测走同一条 fail-open 路径。

**2. Linux 令牌未按启动纪元隔离。** `lt:<ticks-since-boot>` 在 24 小时保留窗口内跨重启可碰撞。现按 `/proc/stat` 的 `btime` 限定。

**3. `toDispatchView` 未扩展 —— 本 PR 的主打修复只落到 3 个消费面中的 1 个。** 账本记下了退出码，但 `/api/dispatch/list` 和 `/api/dispatch/status` 返回的 `DispatchView` 一个退出字段都没有，于是 Web 面板把 `exit 127` 渲染成绿色 "Done"（`lisa-client.ts:3287`），iOS 显示 "Alive: no"（`DispatchLedgerView.swift:61`），与正常完成完全无法区分。

现在 `DispatchView` 带一个派生的 `status`（`running` / `ok` / `failed` / `unknown`）加上 `exitCode` / `exitSignal` / `exitedAt`，OpenAPI 契约同步声明。**`unknown` 刻意区别于 `ok`**：子进程是 detached 的，仅凭 pid，崩溃、被 OOM 杀死和干净退出三者完全一样。

**4. `saveLedger` 是裸 `writeFileSync`。** 它先截断再写，并发读者撞上 `loadLedger()` 的 catch 就静默拿到 `[]`，丢掉全部存活 dispatch。本 PR 的 `recordExit()` 从异步监听器里又加了一次读-改-写，把这个窗口拉得更宽。改为 tmp + rename（`fs-utils.ts` 里 `atomicWrite()` 的同步孪生）。

**5. `_resetIntegrationsForTest()` 是一扇单向门。** built-in 靠 import observer 模块的副作用注册，而 ESM 缓存模块 —— 第二次 `registerBuiltinIntegrations()` 导入十个已求值的模块，一个都注册不上。`hub.test.ts` 目前能过，仅仅因为 reset 恰好是文件里最后执行的东西，**而本 PR 正好在它后面追加了一个 suite**。改为首次注册时快照、后续调用恢复。

**6. `loadOrchestratorConfig` 是整表替换。** README 让用户打开 `~/.lisa/agents.json` 去开启七个可选 observer 之一，而最自然的手工编辑就是只写那一个 —— 于是三个默认项被静默丢弃：开启 codex 会顺手把 Claude Code、managed、pty 的观察**关掉**，无报错、界面无提示。本 PR 恰好新增了 README 里指向该文件的唯一入口，等于把用户领进这个坑。改为逐键合并；显式 `enabled: false` 仍然覆盖默认。

**7. README 的隐私测试断言。** 把"五个 observer"扩写成"十个"时，括号里那句"每个 adapter 都有隐私测试"被一并拖过了五个它从未覆盖的 adapter。逐个核实：claude-code、codex、opencode、aider、git、shell 六个在测试文件里确有隐私断言；github-pr 和 takoapi 有 observer 测试但无隐私断言；**managed 和 pty 连测试文件都没有**。README 改为据实说"十个里有六个"并点名缺失的四个。

### 复核过、不改的两处

**`isNonCanonicalPath` 的误伤面。** 这道 400 门禁挡在所有路由之前，实测会拒绝裸非 ASCII 路径以及 `{}` `^` `` ` `` `<>` 和空格。docstring 断言"没有合法客户端会发出这种路径" —— 我验证了这个假设：全仓库只有一处把变量插进 API 路径（`/api/room/music/file/${encodeURIComponent(t.id)}`，已编码），iOS 侧插值的 id 是 `p<base36>-<base36>` 这类纯 ASCII。当前无误伤。但这是一个**依赖调用方自律的假设**，不是结构性保证：将来任何一处把用户可控字符串（KB 文档标题、soul 名字）拼进路径而忘了编码，都会变成一个静默的 400。值得在新增路由时留意。

**`pending.exit` / `ledgerId` 的竞态。** 逐路径检查过：close 先于账本写入时 `saveExit` 空转但结果已暂存，`ledgerId` 赋值后补写；close 后到时直接写。Node 单线程保证了 `recordDispatch` 的同步读-改-写不会被监听器插入。正确。

### 仍未处理（本 PR 不做，理由如下）

- **Web 面板与 iOS 对新 `status` 字段的消费。** 服务端契约在本 PR 落地，**UI 消费留给各自的表面 PR**：#375 把 `lisa-client.ts` 整体搬去 `src/web/assets/client/main.js`（-4014/+5054），现在改老文件是纯废工且必然冲突；iOS 那处交给 #371。我会在评审这两个 PR 时逐个兑现，不会掉。
- **no-telemetry 守卫的三个洞** 与 **`SuggestionCategory` / `dispatch_status` 输出串** 这两项待你拍板的问题 —— 前者本 PR 的 `de3d76f` 已覆盖目录扫描面，后两者是产品决策不是缺陷。

### 验证

`npm run typecheck` 通过；`npm test` **1699 个，1698 通过，0 失败，1 跳过**（PTY，Node 26 本机已知）；`npm run check:api-contract` 通过。

每条修复都带一个回退即红的测试。以 TZ 那条为例，单独撤掉环境钉死：

```
--- with env pin REVERTED ---
✖ the token does not move when the caller's TZ/locale does
ℹ pass 36  ℹ fail 1
--- restored ---
ℹ pass 37  ℹ fail 0
```

另外修正了两个此前写死 `ps:` 前缀的既有测试 —— 写死前缀在 Linux 上（令牌一直是 `lt:`）本来就是靠巧合通过的，现在从真实探测里取。

